PR #41, opened 2019-07-24
Add support for Mentor Questa simulator, fix seed and other warnings

vandanaprabhu:
We noticed that the seed was not working for Xcelium so we fixed that

We got new compile and simulation warnings when we ran on Questa so we fixed those

We are unable to test these changes on VCS but we made sure that Xcelium still works

Please see individual commit messages for details of the changes made

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ processor verification. It currently supports the following features:
 
 To be able to run the instruction generator, you need to have an RTL simulator
 which supports Systemverilog and UVM 1.2. This generator has been verified with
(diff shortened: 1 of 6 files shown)


vandanaprabhu:
We would like to mention here that we have observed Mentor's Questa runs 2x faster than Cadence Xcelium in our environment when using `./run -test all` option which makes a case for us to go with the Mentor flow for now. We don't have VCS available for comparing. 
In wall clock time we see Mentor takes 22 minutes vs 51 mins on Cadence.

PR #43, opened 2019-07-25
CSR Generation Script and YAML template

udinator:


run
```diff
@@ -37,6 +37,15 @@ SEED=`date +%s`
 # Test name, "all" means run all tests in the testlist
 TEST="riscv_instr_base_test"
```
taoliug: Please keep this file local, I'd expect to roll out a new flow soon.

scripts/gen_csr_test.py
```diff
@@ -0,0 +1,220 @@
...
+import yaml
+import argparse
+import random
+from bitstring import BitArray as bitarray
+# To install the bitstring library:
+#   1) sudo apt-get install python3-bitstring OR
+#   2) pip install bitstring
+
+
+"""
```
taoliug: Please follow the docstring style guide here
http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

scripts/gen_csr_test.py
```diff
@@ -0,0 +1,220 @@
...
+              csr_test_file.write(final_branch_check)
+      # Add final signatures to test, indicating to testbench that test has ended
+      gen_csr_test_pass(csr_test_file)
+      gen_csr_test_fail(csr_test_file)
+
+
+# Add argument options
+parser = argparse.ArgumentParser()
+parser.add_argument("--csr_file", type=str, help="The YAML file contating descriptions of all processor supported CSRs")
+parser.add_argument("--rv", type=int, default=32, help="Specify the ISA width, e.g. 32 or 64 or 128")
```
taoliug: rename this to xlen

scripts/gen_csr_test.py
```diff
@@ -0,0 +1,220 @@
...
+parser.add_argument("--num_test", type=int, default=1, help="Specify how many tests to be generated")
```
taoliug: Looks like this is not used

scripts/gen_csr_test.py
```diff
@@ -0,0 +1,220 @@
...
+
+import sys
+import yaml
+import argparse
+import random
+from bitstring import BitArray as bitarray
+# To install the bitstring library:
+#   1) sudo apt-get install python3-bitstring OR
+#   2) pip install bitstring
+
```
taoliug: Add some description about the CSR test procedure

scripts/gen_csr_test.py
udinator: num_test is used by gen_csr_inst() to specify how many output files to generate

scripts/gen_csr_test.py
udinator: Done

yaml/csr_template.yaml
```diff
@@ -0,0 +1,72 @@
...
+        Encodes native base ISA width
+      type: WARL
+      reset_val: 1
+      msb: 31
+      lsb: 30
+    - field_name: Extensions
+      description: >
+          Encodes all supported ISA extensions
+      type: WARL
+      reset_val: 260 # calculate based on the encoding table in the RISC-V Privileged Spec
```
taoliug: Is this a decimal or hex value? I think we should mandate using hex value like 0x...

scottj97:
Can someone explain what this is all about? Does it have anything to do with the SystemVerilog test generator?

scripts/gen_csr_test.py
udinator: Done

yaml/csr_template.yaml
udinator: Agreed. Changed to hex format.

udinator:
@scottj97, this is essentially an extension for the SV test generator, that allows the capability to generate self-checking tests for CSRs, based off of a YAML description of all processor supported CSRs that the designer provides (as shown in csr_template.yaml).

scottj97:
Right, but I don't see how it is an "extension" of anything...it looks like a completely separate tool. Does it share anything with the code that's already here? It looks like it could be its own separate repo.

taoliug:
> Right, but I don't see how it is an "extension" of anything...it looks like a completely separate tool. Does it share anything with the code that's already here? It looks like it could be its own separate repo.

Hi Scott, we have internally discussed about it and believe this should be an addition of current instruction generator(even though currently it's not sharing anything). The CSR YAML could be used by the SV instruction generator later.

PR #63, opened 2019-08-03
Add main entry point for run.py

danghai:
Hi, I make a PR to minor modify the run.py script. Could you review it when you have the chance? Thank you

### DIFF
--- a/run.py
+++ b/run.py
@@ -262,56 +262,63 @@ def iss_cmp(test_list, iss, output_dir, isa, verbose):
   run_cmd(("echo %s >> %s" % (summary, report)))
   print("ISS regression report is saved to %s" % report)
 
+def setup_parser():
+  """Create a command line parser.
 
-# Parse input arguments
-parser = argparse.ArgumentParser()
+  Returns: The created parser.
+  """
+  # Parse input arguments
+  parser = argparse.ArgumentParser()
 
-parser.add_argument("--o", type=str, default="./out",
-                    help="Output directory name")
-parser.add_argument("--testlist", type=str, default="",
-                    help="Regression testlist")
-parser.add_argument("--isa", type=str, default="rv64imc",
-                    help="RISC-V ISA subset")
-parser.add_argument("--mabi", type=str, default="lp64",
-                    help="mabi used for compilation, lp32 or lp64")
-parser.add_argument("--test", type=str, default="all",
-                    help="Test name, 'all' means all tests in the list")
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -262,94 +263,108 @@ def iss_cmp(test_list, iss, output_dir, isa, verbose):
...
+    args.iss_yaml = cwd + "/yaml/iss.yaml"
+
+  if not args.simulator_yaml:
+    args.simulator_yaml = cwd + "/yaml/simulator.yaml"
+
+  if not args.testlist:
+    args.testlist = cwd + "/yaml/testlist.yaml"
+
+  # Create output directory
+  output_dir = args.o + "_" + str(date.today())
```
taoliug: Can you add date suffix only when args.o == default value? I do want to user be able to specify the exact output directory.

danghai: @taoliug I resolved it.

run.py
```diff
@@ -262,94 +263,111 @@ def iss_cmp(test_list, iss, output_dir, isa, verbose):
...
-  # Compare ISS simulation result
-  if args.steps == "all" or re.match("iss_cmp", args.steps):
-    iss_cmp(matched_list, args.iss, args.o, args.isa, args.verbose)
+  Returns: The created parser.
+  """
+  # Parse input arguments
+  parser = argparse.ArgumentParser()
+
+  parser.add_argument("--o", type=str,
+                      help="Output directory name")
```
taoliug: I think this will trigger an error is user doesn't pass "--o xxx"

danghai: For default, `output_dir = "out_" + str(date.today())`

taoliug: I mean --o becomes a mandatory argument now, there's no way user can run this script without specifying --o.

danghai: I can run this script without specifying --o.
```
danghai@ubuntu:~/google/riscv-dv$ python3 run.py --test riscv_arithmetic_basic_test --simulator questa
Processing regression test list : /home/danghai/google/riscv-dv/yaml/testlist.yaml, test: riscv_arithmetic_basic_test
Found matched tests: riscv_arithmetic_basic_test, iterations:2
Processing simulator setup file : /home/danghai/google/riscv-dv/yaml/simulator.yaml
Found matching simulator: questa
Building RISC-V instruction generator
Running RISC-V instruction generator
Generating 2 riscv_arithmetic_basic_test
```

Output: `out_2019-08-03`

PR #66, opened 2019-08-08
CSR integration, general simulation correctness report

udinator:


run.py
```diff
@@ -263,10 +279,12 @@ def setup_parser():
   # Parse input arguments
   parser = argparse.ArgumentParser()
 
-  parser.add_argument("--o", type=str,
+  parser.add_argument("--o", type=str, default="./out",
                       help="Output directory name")
```
taoliug: looks like you are overriding the previous change?

run.py
```diff
@@ -285,7 +303,7 @@ def setup_parser():
                       help="RISC-V instruction set simulator: spike, ovpsim")
   parser.add_argument("--iss_yaml", type=str, default="",
                       help="ISS setting YAML")
-  parser.add_argument("-v", "--verbose", dest="verbose", action="store_true",
+  parser.add_argument("--verbose", dest="verbose", type=int,
```
taoliug: Same here

run.py
```diff
@@ -344,12 +363,8 @@ def main():
     args.testlist = cwd + "/yaml/testlist.yaml"
 
   # Create output directory
-  if args.o is None:
-    output_dir = "out_" + str(date.today())
-  else:
-    output_dir = args.o
-  subprocess.run(["mkdir", "-p", output_dir])
-  subprocess.run(["mkdir", "-p", ("%s/asm_tests" % output_dir)])
+  subprocess.run(["mkdir", "-p", args.o])
```
taoliug: Same here

yaml/testlist.yaml
```diff
@@ -76,23 +85,27 @@
...
     +directed_instr_0=riscv_load_store_rand_instr_stream,8
   rtl_test: core_base_test
+  no_post_compare: 0
 
 - test: riscv_mmu_stress_test
   description: >
     Test with different patterns of load/store instructions, stress test MMU
     operations.
   iterations: 2
+  no_iss: 0
```
taoliug: can you make no_iss and no_compare optional? For most tests, this doesn't need to be specified. Maybe just need to modify the parsing script a little bit.

run.py
```diff
@@ -137,21 +140,32 @@ def gen(test_list, simulator, simulator_yaml, output_dir, sim_only,
...
-        cmd = lsf_cmd + " " + sim_cmd.rstrip() + \
-              (" +UVM_TESTNAME=%s " % test['gen_test']) + \
-              (" +num_of_tests=%d " % test['iterations']) + \
-              (" +asm_file_name=%s/asm_tests/%s " % (output_dir, test['test'])) + \
-              (" -l %s/sim_%s.log " % (output_dir, test['test']))
-        cmd = re.sub("<seed>", str(rand_seed), cmd)
-        if "gen_opts" in test:
-          cmd += test['gen_opts']
-        logging.info("Generating %d %s" % (test['iterations'], test['test']))
+      if test['test'] == 'riscv_csr_test' and test['iterations'] > 0:
```
taoliug: Add some comment here why CSR test needs a different command

yaml/csr_template.yaml
```diff
@@ -53,7 +55,9 @@
 - csr: misa
   description: >
     Machine ISA Register
+  address: 0x301
   privilege_mode: M
+  read_only: 0
```
taoliug: Should this be applied to each field rather the whole register? BTW, does this mean the "type" is now redundant or inaccurate?

yaml/testlist.yaml
udinator: sure, will do that

run.py
udinator: done

yaml/csr_template.yaml
udinator: good point, there is also this configurability for each field as well, so the CSR level read_only is not needed.

yaml/testlist.yaml
```diff
@@ -205,3 +207,11 @@
     +enable_irq_seq=1
   compare_opts: >
     +compare_final_value_only=1
+
+- test: riscv_csr_test
+  description: >
+    Test all CSR instructions on all implemented CSR registers
+  iterations: 5
```
taoliug: Can you change it to 1 iteration by default?
BTW, we should update the README.md as the CSR script requires additional package to run.

run.py
```diff
@@ -205,14 +223,17 @@ def iss_sim(test_list, output_dir, iss_list, iss_yaml, isa, timeout_s):
...
-        logging.debug(cmd)
+      if 'no_iss' in test and test['no_iss'] == 1:
+        continue
+      else:
+        for i in range(0, test['iterations']):
+          prefix = ("%s/asm_tests/%s.%d" % (output_dir, test['test'], i))
+          elf = prefix + ".o"
+          log = ("%s/%s.%d.log" % (log_dir, test['test'], i))
+          cmd = get_iss_cmd(base_cmd, elf, log)
+          print ("Running ISS simulation: %s" % elf)
```
danghai: I think you should use `logging.info` instead of print

yaml/csr_template.yaml
```diff
@@ -36,7 +37,7 @@
 #    - MSB_FIELD_NAME:
 #      - description: >
 #          BRIEF_DESCRIPTION
-#      - type: TYPE (WPRI/WLRL/WARL)
+#      - type: TYPE (WPRI/WLRL/WARL/R)
```
taoliug: I think you are using mixed terminologies here. We should probably change to standard CSR terminology like WR/RO/W0C/W1C etc. And for WR field, we should specify the allowed values, rather than WARL which doesn't tell too much what's allowed. The allowed values could be done in a later PR.

run.py
```diff
@@ -99,12 +99,14 @@ def get_iss_cmd(base_cmd, elf, log):
...
-def gen(test_list, simulator, simulator_yaml, output_dir, sim_only,
-        compile_only, lsf_cmd, seed, cwd, cmp_opts, sim_opts, timeout_s):
+def gen(test_list, csr_file, isa, simulator, simulator_yaml, output_dir, sim_only,
+        compile_only, lsf_cmd, seed, cwd, cmp_opts, sim_opts, iterations, timeout_s):
   """Run the instruction generator
 
   Args:
     test_list      : List of assembly programs to be compiled
+    csr_file       : YAML file containing description of all CSRs
+    isa            : Processor supported ISA subset
```
danghai: It misses document for `iterations`, as well as `parser.add_argument` for this option?

run.py
udinator: Sure, will change this

udinator: Sure, will change this

udinator: Sure, will change this

run.py
udinator: The add_argument() call exists, but you're right, the documentation is not. Thanks for pointing that out, will add that in.

yaml/csr_template.yaml
udinator: marking resolved for now

run.py
```diff
@@ -137,21 +141,36 @@ def gen(test_list, simulator, simulator_yaml, output_dir, sim_only,
...
-              (" -l %s/sim_%s.log " % (output_dir, test['test']))
-        cmd = re.sub("<seed>", str(rand_seed), cmd)
-        if "gen_opts" in test:
-          cmd += test['gen_opts']
-        logging.info("Generating %d %s" % (test['iterations'], test['test']))
+      """
+      If we are running a CSR test, need to call a separate python script
+      to generate directed CSR test code, located at scripts/gen_csr_test.py.
+      """
+      if test['test'] == 'riscv_csr_test' and test['iterations'] > 0:
```
danghai: In my opinion, I think you should change the flow `if` condition. I mean:
```
if test[`iteration`] > 0:
     if test[`test`] == `riscv_csr_test`:
          ....
          ....
     else:
          ....
          ....
     if lsf_cmd:
          cmd_list.append(cmd)
     else:
          run_cmd(cmd, timeout_s)
```

What do you think?

udinator: yes, you're right, that would be more streamlined logic.

run.py
```diff
@@ -138,15 +142,26 @@ def gen(test_list, simulator, simulator_yaml, output_dir, sim_only,
...
-              (" -l %s/sim_%s.log " % (output_dir, test['test']))
-        cmd = re.sub("<seed>", str(rand_seed), cmd)
-        if "gen_opts" in test:
-          cmd += test['gen_opts']
+        """
+        If we are running a CSR test, need to call a separate python script
+        to generate directed CSR test code, located at scripts/gen_csr_test.py.
+        """
+        if test['test'] == 'riscv_csr_test':
+          cmd = "python3 scripts/gen_csr_test.py" + \
```
taoliug: Does this script accept seed? It could be added later.

run.py
```diff
@@ -99,12 +99,14 @@ def get_iss_cmd(base_cmd, elf, log):
   return cmd
 
 
-def gen(test_list, simulator, simulator_yaml, output_dir, sim_only,
-        compile_only, lsf_cmd, seed, cwd, cmp_opts, sim_opts, timeout_s):
+def gen(test_list, csr_file, isa, simulator, simulator_yaml, output_dir, sim_only,
+        compile_only, lsf_cmd, seed, cwd, cmp_opts, sim_opts, iterations, timeout_s):
```
taoliug: Why not use test['iterations']?

README.md
```diff
@@ -84,22 +97,104 @@ python3 run.py --test riscv_page_table_exception_test --co
...
+Any number of the above options can be combined together, and all of them have
+default values should they not be specified at runtime.
+
 ## Configuration
 
+### Setup CSR description list
+
+[CSR descriptions in YAML
+format](https://github.com/google/riscv-dv/blob/master/yaml/csr_template.yaml))
+
```
taoliug: Unmatched ) at the end

README.md
```diff
@@ -84,22 +97,104 @@ python3 run.py --test riscv_page_table_exception_test --co
...
+```
+
+Any number of the above options can be combined together, and all of them have
+default values should they not be specified at runtime.
+
 ## Configuration
 
+### Setup CSR description list
+
+[CSR descriptions in YAML
```
taoliug: Please add a description - "This is only needed for the privileged CSR test"

run.py
udinator: The gen_csr script doesn't use a seed right now, but it is something that I could add in.

README.md
```diff
@@ -84,22 +97,104 @@ python3 run.py --test riscv_page_table_exception_test --co
 ....
 ```
 
+### CSR Generation
```
taoliug: Privileged CSR test generation

README.md
```diff
@@ -84,22 +97,104 @@ python3 run.py --test riscv_page_table_exception_test --co
...
+
+
+The CSR generation script is located at [scripts/gen_csr_test.py](https://github.com/google/riscv-dv/blob/master/scripts/gen_csr_test.py).  To quickly generate a basic CSR test, run the below command:
```
taoliug: Please add some description about the intent and procedure of the privileged CSR test.

run.py
udinator: yes, the iterations argument is redundant since yaml processing has already happened

PR #68, opened 2019-08-08
Give error for mutually exclusive between -co, and -so argument

danghai:
I make PR to give error for mutually exclusive between `-co`, and `-so` argument


taoliug:
Thanks for adding this

PR #70, opened 2019-08-12
add pass_val and fail_val into csr test flow for EOT correctness check

udinator:


run.py
```diff
@@ -290,6 +297,12 @@ def setup_parser():
                       help="Regression testlist")
   parser.add_argument("--csr_yaml", type=str, default="",
                       help="CSR description file")
+  parser.add_argument("--pass_val", type=int, default=1,
+                      help="Value that privileged CSR test writes on success")
+  parser.add_argument("--fail_val", type=int, default=2,
```
taoliug: Use 0 for failure?

scripts/gen_csr_test.py
```diff
@@ -312,6 +321,12 @@ def gen_csr_instr(csr_map, csr_instructions, xlen, iterations, out):
         help="Specify how many tests to be generated")
 parser.add_argument("--out", type=str, default="./",
         help="Specify output directory")
+parser.add_argument("--pass_val", type=int, default=0,
```
taoliug: Use the same value as the run script? Actually I feel that there's no need to add this argument. PASS = 1, FAIL = 0 seems to be good enough.  This could simplify the script interface.

run.py
```diff
@@ -99,13 +99,17 @@ def get_iss_cmd(base_cmd, elf, log):
...
+        simulator_yaml, output_dir, sim_only, compile_only, lsf_cmd, seed,
+        cwd, cmp_opts, sim_opts, timeout_s):
   """Run the instruction generator
 
   Args:
     test_list      : List of assembly programs to be compiled
     csr_file       : YAML file containing description of all CSRs
+    pass_val       : Value that priv. CSR test writes on success
+    fail_val       : Value that priv. CSR test writes on failure
+    csr_end_addr   : Address that priv. CSR test writes to at end of test
```
taoliug: maybe change it to signature_addr? I think this could be used by other tests later.

run.py
udinator: Sure, can switch to zero. I initially wanted to avoid any false positive EOT checks since 0 is a common default value

run.py
udinator: Will do.

scripts/gen_csr_test.py
udinator: Sure

run.py
```diff
@@ -99,24 +99,26 @@ def get_iss_cmd(base_cmd, elf, log):
...
-    sim_only       : Simulation only
-    compile_only   : Compile the generator only
-    lsf_cmd        : LSF command used to run the instruction generator
-    seed           : Seed to the instruction generator
-    cmp_opts       : Compile options for the generator
-    sim_opts       : Simulation options for the generator
-    timeout_s      : Timeout limit in seconds
+    test_list             : List of assembly programs to be compiled
+    csr_file              : YAML file containing description of all CSRs
+    end_signature_addr    : Address that priv. CSR test writes to at end of test
```
taoliug: Please modify the description of the argument to make it more general. It should be used for writing pass/fail signature in general.

udinator: Done

PR #78, opened 2019-08-15
Add basic debug test functionality

udinator:
- Modify instruction generator to generate a very basic debug rom, contains random instruction stream
- Modify trace comparison script to change outdated primitive types

### DIFF
--- a/scripts/instr_trace_compare.py
+++ b/scripts/instr_trace_compare.py
@@ -32,6 +32,10 @@ def compare_trace_csv(csv1, csv2, name1, name2, log,
   matched_cnt = 0
   mismatch_cnt = 0
 
+  # ensure that in order mode is disabled if necessary
+  if compare_final_value_only:
+    in_order_mode = 0
+
   if log:
     fd = open(log, 'a+')
   else:
@@ -131,8 +135,8 @@ def compare_trace_csv(csv1, csv2, name1, name2, log,
           for trace_1_index in range(0, len(gpr_trace_1[gpr])-1):
             if (trace_2_index == len(gpr_trace_2[gpr])):
               break
-            if long(gpr_trace_1[gpr][trace_1_index].rd_val, 16) != \
-               long(gpr_trace_2[gpr][trace_2_index].rd_val, 16):
+            if int(gpr_trace_1[gpr][trace_1_index].rd_val, 16) != \
+               int(gpr_trace_2[gpr][trace_2_index].rd_val, 16):
               if coalesced_updates >= coalescing_limit:
                 coalesced_updates = 0
                 mismatch_cnt += 1
@@ -163,8 +167,8 @@ def compare_trace_csv(csv1, csv2, name1, name2, log,
           mismatch_cnt += 1
           fd.write("Zero GPR[%s] updates observed: %s:%d, %s:%d\n" % (gpr,
                    name1, len(gpr_trace_1[gpr]), name2, len(gpr_trace_2[gpr])))
-        elif long(gpr_trace_1[gpr][-1].rd_val, 16) != \
-             long(gpr_trace_2[gpr][-1].rd_val, 16):
+        elif int(gpr_trace_1[gpr][-1].rd_val, 16) != \
+             int(gpr_trace_2[gpr][-1].rd_val, 16):
           mismatch_cnt += 1
           if mismatch_cnt <= mismatch_print_limit:
             fd.write("Mismatch final value:\n")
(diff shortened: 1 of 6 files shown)


scripts/instr_trace_compare.py
```diff
@@ -131,8 +135,8 @@ def compare_trace_csv(csv1, csv2, name1, name2, log,
           for trace_1_index in range(0, len(gpr_trace_1[gpr])-1):
             if (trace_2_index == len(gpr_trace_2[gpr])):
               break
-            if long(gpr_trace_1[gpr][trace_1_index].rd_val, 16) != \
-               long(gpr_trace_2[gpr][trace_2_index].rd_val, 16):
+            if int(gpr_trace_1[gpr][trace_1_index].rd_val, 16) != \
```
taoliug: Is "long" causing problem here? I am not sure if your change will work for XLEN = 64 case

src/riscv_asm_program_gen.sv
```diff
@@ -918,13 +925,37 @@ class riscv_asm_program_gen extends uvm_object;
...
-      gen_section("debug_rom", instr);
-      instr = {"dret"};
-      gen_section("debug_exception", instr);
-    end
+      // The main debug rom
+      debug_program = riscv_instr_sequence::type_id::create("debug_program");
+      debug_program.instr_cnt = cfg.debug_program_instr_cnt;
+      debug_program.is_debug_program = 1;
+      debug_program.cfg = cfg;
+      //debug_program.label_name = "debug_rom";
```
taoliug: Remove temporarily code

src/riscv_asm_program_gen.sv
```diff
@@ -918,13 +925,37 @@ class riscv_asm_program_gen extends uvm_object;
...
+      `DV_CHECK_RANDOMIZE_FATAL(debug_program)
+      debug_program.gen_instr(.is_main_program(1'b1), .enable_hint_instr(1'b0),
+                              .no_branch(1'b1));
```
taoliug: is this temporarily? I'd expect to test branch instruction in debug rom later.

src/riscv_asm_program_gen.sv
```diff
@@ -918,13 +925,37 @@ class riscv_asm_program_gen extends uvm_object;
...
+      debug_program.post_process_instr();
+      debug_program.generate_instr_stream(.no_label(1'b1));
+      // Need to save off GPRs to avoid modifying program flow
+      push_gpr_to_kernel_stack(MSTATUS, MSCRATCH, cfg.mstatus_mprv, push_gpr);
+      gen_section("debug_rom", push_gpr);
+      dret = debug_program.instr_string_list.pop_back();
+      instr_stream = {instr_stream, debug_program.instr_string_list};
+      pop_gpr_from_kernel_stack(MSTATUS, MSCRATCH, cfg.mstatus_mprv, pop_gpr);
+      gen_section("", pop_gpr);
+      instr_stream = {instr_stream, dret};
```
taoliug: I think we'd better to have a mode only do dret so that we can compare each instruction with ISS.

src/riscv_asm_program_gen.sv
taoliug: Maybe add a TODO here

scripts/instr_trace_compare.py
udinator: yes, using "long" is causing python3 to error out. It looks like the "long" and "int" types were unified in Python 3.x

src/riscv_asm_program_gen.sv
udinator: Yes, this is temporary, just for the basic debug test. Once I add branches in, then there will be some sort of configurability between branches being enabled/disabled

src/riscv_asm_program_gen.sv
udinator: Done.

src/riscv_instr_gen_config.sv
```diff
@@ -62,6 +65,13 @@ class riscv_instr_gen_config extends uvm_object;
   bit                    check_misa_init_val = 1'b0;
   bit                    check_xstatus = 1'b1;
 
+  // Enable a full or empty debug_rom section.
+  // Full debug_rom will contain random instruction streams.
+  // Empty debug_rom will contain just dret instruction
+  //   and will return immediately.
+  // Will be empty by default.
+  bit               empty_debug_section = 1'b0;
```
taoliug: Alignment is off. Besides, please use full line length (100) for the comment

src/riscv_instr_sequence.sv
```diff
@@ -267,6 +274,10 @@ class riscv_instr_sequence extends uvm_sequence;
       str = {prefix, "ret"};
       instr_string_list.push_back(str);
     end
+    if (is_debug_program) begin
```
taoliug: is this duplicated?

udinator: No, this check only takes place here in the instr_sequence to determine whether to add a ret or dret instruction at the end of the sequence

src/riscv_instr_gen_config.sv

PR #79, opened 2019-08-15
Add shorten option

danghai:
I make PR to add shorten option.

### DIFF
--- a/run.py
+++ b/run.py
@@ -293,9 +293,9 @@ def setup_parser():
   # Parse input arguments
   parser = argparse.ArgumentParser()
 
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -293,25 +293,25 @@ def setup_parser():
   # Parse input arguments
   parser = argparse.ArgumentParser()
 
-  parser.add_argument("--o", type=str,
+  parser.add_argument("-o", "--ouput", type=str,
```
scottj97: Output is misspelled.

taoliug: This is now fixed in #80

PR #84, opened 2019-08-20
Fix license URLs in comments

imphil:
It seems a global search-replace broke these URLs, fixing them.


imphil:
@googlebot I signed it!

PR #91, opened 2019-08-22
Initial signature enum for handshake protocol

udinator:


src/riscv_instr_pkg.sv
```diff
@@ -26,6 +26,43 @@ package riscv_instr_pkg;
 
   parameter CORE_INITIALIZATION_DONE = 2;
 
+  typedef enum bit[XLEN-1:0] {
```
taoliug: I think it's better to leave a few bits for other purpose. Maybe 8 bits are good enough for the signature type?

udinator: I think you're right, that would free up the rest of the data word, and then I could potentially combine the CSR address and data for the WRITE_CSR into one data word to keep the number of handshake writes lower

src/riscv_instr_pkg.sv
```diff
@@ -26,6 +26,43 @@ package riscv_instr_pkg;
...
+
+  typedef enum bit {
+    CORE_NOT_READY,
+    CORE_READY
+  } core_start_t;
+
+  typedef enum bit {
+    IRQ_REQ,
+    DEBUG_REQ
+  } send_req_t;
```
taoliug: How does the instruction generator use this enum?

udinator: My thinking was that for situations where directed stimulation at a particular time is necessary, such as a debug request during an interrupt handler, or something like that, this enum would be useful to allow this

src/riscv_instr_pkg.sv
```diff
@@ -26,6 +26,43 @@ package riscv_instr_pkg;
...
   parameter CORE_INITIALIZATION_DONE = 2;
 
+  typedef enum bit[XLEN-1:0] {
+    // Information sent to the core relating to core initialization
+    // Will be followed by information from core_start_t.
+    CORE_START,
+    // Information sent to the core relating to when to send external stimulus
+    // (such as debug or interrupt stimulus).
+    // Will be followed by information from send_req_t.
+    SEND_REQ,
```
taoliug: What's the difference between SEND_REQ and CORE_START?

udinator: CORE_START is just to signal the testbench that it is okay to start sending stimulus in general, no matter at what point stimulus is sent, but SEND_REQ is to indicate that stimulus should be sent at that particular moment, for more directed tests

src/riscv_instr_pkg.sv
taoliug: My thought is that the signature should be used to indicate the status of the core. For example, you can rename the CORE_START to CORE_STATUS, and you can use it to indicate the core is initialized, core is in debug mode etc. I don't want to it to be too specific to the external req

src/riscv_instr_pkg.sv
taoliug: Please refer to the comment, I think these two can be merged to one.

src/riscv_instr_pkg.sv
udinator: Agreed, it has been changed to CORE_STATUS, with some relevant options

PR #156, opened 2019-09-13
Interrupt test integration

udinator:
-Add handshakes to write xIE, xIP, and xCAUSE to the testbench on exceptions and interrupts for more checking analysis
- Removed all csr WIRI fields as per the spec
- make WPRI csr fields writable, to allow for platform-specific csr configurations
- randomize csr generation, in particular the interrupt enable bits in xSTATUS and xIE

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -478,6 +478,8 @@ class riscv_asm_program_gen extends uvm_object;
     privil_seq = riscv_privileged_common_seq::type_id::create("privil_seq");
     foreach(riscv_instr_pkg::supported_privileged_mode[i]) begin
       string instr[$];
+      string csr_handshake[$];
+      string ret_instr;
       if(riscv_instr_pkg::supported_privileged_mode[i] < cfg.init_privileged_mode) continue;
       `uvm_info(`gfn, $sformatf("Generating privileged mode routing for %0s",
                       riscv_instr_pkg::supported_privileged_mode[i].name()), UVM_LOW)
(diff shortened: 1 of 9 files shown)


src/riscv_asm_program_gen.sv
```diff
@@ -719,22 +745,27 @@ class riscv_asm_program_gen extends uvm_object;
     instr = {instr, ".option norvc;",
                     $sformatf("j %0smode_exception_handler", mode)};
     // Redirect the interrupt to the corresponding interrupt handler
-    for (int i = 1; i < 16; i++) begin
+    for (int i = 1; i < 32; i++) begin
```
taoliug: Can you  make this a config parameter. Based on the privileged spec, i >= 12 is reserved.

src/riscv_asm_program_gen.sv
```diff
@@ -719,22 +745,27 @@ class riscv_asm_program_gen extends uvm_object;
...
       instr.push_back($sformatf("j %0smode_intr_vector_%0d", mode, i));
     end
     instr = {instr, ".option rvc;"};
-    for (int i = 1; i < 16; i++) begin
+    for (int i = 1; i < 32; i++) begin
       string intr_handler[$];
       push_gpr_to_kernel_stack(status, scratch, cfg.mstatus_mprv, intr_handler);
-      gen_signature_handshake(intr_handler, CORE_STATUS, HANDLING_IRQ);
+      gen_signature_handshake(.instr(intr_handler), .signature_type(CORE_STATUS), .core_status(HANDLING_IRQ));
```
taoliug: Should this be inside if (cfg.require_handshake)?

taoliug: You could potentially do it inside gen_signature_handshake function so that you don't need to check it everywhere

udinator: No, as the gen_signature_handshake() function checks for cfg.require_signature_addr before doing anything else, just to avoid having to write this if-condition everywhere throughout riscv_asm_gen

src/riscv_instr_gen_config.sv
```diff
@@ -57,6 +57,10 @@ class riscv_instr_gen_config extends uvm_object;
   rand bit               mstatus_tvm;
   rand mtvec_mode_t      mtvec_mode;
 
+  rand bit[XLEN-1:0] mstatus, mie,
+                     sstatus, sie,
```
taoliug: There are some random bits for mstatus already, how does the new rand variable work with the existing ones? Can you combine them?

src/riscv_privileged_common_seq.sv
```diff
@@ -132,14 +153,22 @@ class riscv_privileged_common_seq extends uvm_sequence;
...
+      ustatus.set_field("UIE", riscv_instr_pkg::support_umode_trap);
+    end
     regs.push_back(ustatus);
     if (UIE inside {implemented_csr}) begin
       uie = riscv_privil_reg::type_id::create("uie");
       uie.init_reg(UIE);
-      uie.set_field("UEIE", cfg.enable_interrupt);
-      uie.set_field("USIE", cfg.enable_interrupt);
+      uie.set_val(cfg.uie);
+      if (!cfg.enable_interrupt) begin
```
taoliug: why not just constraint uie with enable_interrupt? so that you don't need to do field based setting?

src/riscv_asm_program_gen.sv
udinator: Yep, will do, I'll add this to riscv_core_setting.sv, as this feature needs to remain static on a core-by-core basis.

src/riscv_instr_gen_config.sv
udinator: Currently, what i do is overwrite the fields after csr randomization with these random bits, but you're right they can be combined. Easiest thing to do is probably assign these named bit variables to the appropriate bits within `rand mstatus` and so on.

src/riscv_privileged_common_seq.sv
udinator: Done.

src/riscv_instr_gen_config.sv

src/riscv_asm_program_gen.sv
```diff
@@ -719,22 +745,27 @@ class riscv_asm_program_gen extends uvm_object;
     instr = {instr, ".option norvc;",
                     $sformatf("j %0smode_exception_handler", mode)};
     // Redirect the interrupt to the corresponding interrupt handler
-    for (int i = 1; i < 16; i++) begin
+    for (int i = 1; i < max_interrupt_vector_num; i++) begin
       instr.push_back($sformatf("j %0smode_intr_vector_%0d", mode, i));
     end
     instr = {instr, ".option rvc;"};
-    for (int i = 1; i < 16; i++) begin
+    for (int i = 1; i < 32; i++) begin
```
taoliug: Should this 32 be changed as well?

udinator: yes, you're right, has been done

PR #159, opened 2019-09-16
Adding dsim support

taoliug:


src/riscv_instr_pkg.sv
```diff
@@ -835,8 +831,4 @@ package riscv_instr_pkg;
   `include "riscv_instr_sequence.sv"
   `include "riscv_asm_program_gen.sv"
 
-  `ifdef RISCV_DV_EXT_FILE_LIST
-    `include_file(`RISCV_DV_EXT_FILE_LIST)
-  `endif
```
sdexter-numa: README states:

Do not modify the upstream classes directly.
When possible, extending from the upstream classses and implment your own functionalities.
Use command line type override to use your extended classes.
--sim_opts="+uvm_set_type_override=<upstream_class>,<extended_class>"
Create a new file list for your local modifications. Pass to the instruction generator like below:
--cmp_opts "+define+RISCV_DV_EXT_FILE_LIST=<local_file_list>"


How is this possible if these lines have been removed? Is there an alternative that is replacing this?

taoliug: updated README, please take a look.

sdexter-numa: Apologies, but unfortunately that doesn't explain or solve the issue. It is possible to have a custom version of the riscv_core_setting.sv, but because of the removal of...

> `ifdef RISCV_DV_EXT_FILE_LIST
>   `include_file(`RISCV_DV_EXT_FILE_LIST)
> `endif

... it is now not possible to easily create custom versions of other classes. How can I, for example, create and use an extended version of the riscv_instr_base.sv?

taoliug: Please take a look at the updated README
https://github.com/google/riscv-dv#end-to-end-rtl-and-iss-co-simulation-flow

Please add your local extensions under user_extension directory, and included in user_extension.svh file. 

Here's an example:
https://github.com/lowRISC/ibex/tree/master/dv/uvm/riscv_dv_extension

PR #175, opened 2019-09-24
Dump performance counters to testbench at EOT

udinator:


src/riscv_asm_program_gen.sv
```diff
@@ -993,6 +993,11 @@ class riscv_asm_program_gen extends uvm_object;
 
   // Dump performance CSRs if applicable
   virtual function void dump_perf_stats();
+    foreach(implemented_csr[i]) begin
+      if (implemented_csr[i][11:8] === 4'hB) begin
```
taoliug: Can you use inside {....} to make it more readable?

udinator: done.

PR #183, opened 2019-09-26
Fix ovpsim log compare issue

udinator:


udinator:
> never mind, it doesn't matter as you are iterating on gpr_trace_1 already

Yep, plus gpr_trace_1 should in general be the core tracer log, which should show all updates to registers

udinator:
> I think there's a hole that if the GPR exists in trace 2 but not in trace 1, the compare will still pass.

agreed, since the loop is already iterating over gpr_trace_1, adding a second loop that iterates over gpr_trace_2 could work, although it would duplicate the whole block of logic

taoliug:
> > I think there's a hole that if the GPR exists in trace 2 but not in trace 1, the compare will still pass.
> 
> agreed, since the loop is already iterating over gpr_trace_1, adding a second loop that iterates over gpr_trace_2 could work, although it would duplicate the whole block of logic

OK, let's fix this one later

PR #201, opened 2019-10-01
Debug single step functionality and config knobs

udinator:


src/riscv_asm_program_gen.sv
```diff
@@ -1197,17 +1197,56 @@ class riscv_asm_program_gen extends uvm_object;
           // TODO(udinator) - randomize the dcsr.ebreak setup
           gen_dcsr_ebreak(instr);
         end
+        if (cfg.enable_debug_single_step) begin
```
taoliug: Please add a TODO to refactor the whole debug rom generation in a separate class.

udinator: Done.

src/riscv_asm_program_gen.sv
```diff
@@ -1197,17 +1198,56 @@ class riscv_asm_program_gen extends uvm_object;
...
+                 $sformatf("beqz x0, 3f"), // or do nothing and decrement next iteration?
+                 // Decrement dscratch counter
+                 $sformatf("2: csrr x%0d, 0x%0x", cfg.scratch_reg, DSCRATCH0),
+                 $sformatf("addi x%0d, x%0d, -1", cfg.scratch_reg, cfg.scratch_reg),
+                 $sformatf("csrw 0x%0x, x%0d", DSCRATCH0, cfg.scratch_reg),
+                 // Restore scratch_reg value from dscratch1
+                 $sformatf("3: csrr x%0d, 0x%0x", cfg.scratch_reg, DSCRATCH1)
+                 };
+          instr = {instr, str};
+          // write dpc to testbench and make sure it's getting incremented by 0x2 or 0x4
```
taoliug: I think this comment is not accurate because dpc could increase more than this if user are enabling single-stepping with branch/jump. Maybe just say "write DPC to testbench" is enough.

udinator: Good point, comment has been modified.

PR #216, opened 2019-10-15
Added support for coverage of riscvOVPsim

eroom1966:
This pull request includes changes to support riscvOVPsim as a target simulator for coverage analysis from its trace output.
Also includes support for questa in addition to VCS


taoliug:
Can you resolve the conflict? The upstream code has some changes to allow specifying the simulator which conflict with this PR.

PR #229, opened 2019-10-16
Add more ebreak generation control

udinator:
-Enable full configuration of ebreak generation - all permutations of ebreak in main program and debug program are now possible.
-For the debug_ebreak_test in particular, move the DCSR and DPC handshakes within the stack push/pop operations to avoid corrupting program state

### DIFF
--- a/src/riscv_instr_gen_config.sv
+++ b/src/riscv_instr_gen_config.sv
@@ -560,13 +560,11 @@ class riscv_instr_gen_config extends uvm_object;
   virtual function void build_instruction_list();
     basic_instr = {instr_category[SHIFT], instr_category[ARITHMETIC],
                    instr_category[LOGICAL], instr_category[COMPARE]};
-    if (no_ebreak == 0) begin
-      basic_instr = {basic_instr, EBREAK};
-      foreach(riscv_instr_pkg::supported_isa[i]) begin
-        if (riscv_instr_pkg::supported_isa[i] inside {RV32C, RV64C, RV128C, RV32DC, RV32FC}) begin
-          basic_instr = {basic_instr, C_EBREAK};
-          break;
-        end
+    basic_instr = {basic_instr, EBREAK};
+    foreach(riscv_instr_pkg::supported_isa[i]) begin
+      if (riscv_instr_pkg::supported_isa[i] inside {RV32C, RV64C, RV128C, RV32DC, RV32FC}) begin
+        basic_instr = {basic_instr, C_EBREAK};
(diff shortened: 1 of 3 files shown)


src/riscv_instr_gen_config.sv
```diff
@@ -473,7 +473,7 @@ class riscv_instr_gen_config extends uvm_object;
 
   function void post_randomize();
     // Setup the list all reserved registers
-    reserved_regs = {tp, sp, scratch_reg};
+    reserved_regs = {tp, sp, scratch_reg, gpr[1]};
```
taoliug: why reserve gpr[1]?

src/riscv_instr_gen_config.sv
```diff
@@ -560,13 +560,11 @@ class riscv_instr_gen_config extends uvm_object;
   virtual function void build_instruction_list();
     basic_instr = {instr_category[SHIFT], instr_category[ARITHMETIC],
                    instr_category[LOGICAL], instr_category[COMPARE]};
-    if (no_ebreak == 0) begin
```
taoliug: Why remove this condition?

src/riscv_instr_gen_config.sv
udinator: I was seeing that the handshake address register would get used in load operations, creating an invalid memory addres.

src/riscv_instr_gen_config.sv
udinator: Basically, with these changes, cfg.no_ebreak now refers only to the main_program, if it is zero, but cfg.enable_ebreak_in_debug_rom is 1'b1 and a debug program should be generated, then ebreak instructions will be randomly generated inside the debug rom - this logic I placed inside the randomization of riscv_instr_stream.sv

src/riscv_instr_gen_config.sv
taoliug: I think the handshake process is a bunch of directed instructions, how could external load instruction impact the handshake address

udinator: As in, after the handshake has finished, the register containing the signature_addr would get used in other random load instructions

taoliug: Do you initialize the handshake address before you do every handshake?

udinator: yep, it gets loaded every time

udinator: Resolving, as the constraint has now been removed

PR #245, opened 2019-10-20
Fix typo

T3KT4K:
It appears this is just a typo and that the script runs correctly with it fixed


taoliug:
Thanks for reporting, I fixed it with #246

PR #248, opened 2019-10-21
Updating OVPsim device configuration and coverage processing

eroom1966:
Updated OVPsim log processing for coverage
Added OVPsim config files as targets

### DIFF
--- a/scripts/ovpsim_log_to_trace_csv.py
+++ b/scripts/ovpsim_log_to_trace_csv.py
@@ -65,9 +65,13 @@ def process_jalr(trace, operands, gpr):
         trace.rs1_val = gpr[trace.rs1]
         trace.imm = get_imm_hex_val('0')
 
+def process_if_compressed(prev_trace):
+  if len(prev_trace.binary) == 4: # compressed are always 4 hex digits
+    prev_trace.instr = "c."+prev_trace.instr
+    # logging.debug("process_if_compressed(%s, %s)" % (prev_trace.instr, prev_trace.binary))
 
 def process_ovpsim_sim_log(ovpsim_log, csv, full_trace = 0):
-  """Process SPIKE simulation log.
+  """Process OVPsim simulation log.
 
   Extract instruction and affected register information from ovpsim simulation
   log and save to a list.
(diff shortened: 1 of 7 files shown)


target/rv32imc/riscvOVPsim.ic
```diff
@@ -0,0 +1,22 @@
+# riscOVPsim configuration file converted from YAML
+--variant RVB32I
```
taoliug: Does this include RV32 M,C extension?

scripts/ovpsim_log_to_trace_csv.py
```diff
@@ -113,7 +117,7 @@ def process_ovpsim_sim_log(ovpsim_log, csv, full_trace = 0):
         prev_trace.binary = trace_bin
         prev_trace.addr = trace_addr
         #prev_trace.privileged_mode = trace_mode
-        prev_trace.instr = trace_instr
+        prev_trace.instr = trace_instr # wrong
```
taoliug: What's the meaning of this comment?

simon5656: I feel there is some logic wrong in this section and this line '# wrong' might be superflous - but until we have many more regression tests I am just highlighting that this needs looking at in future maybe it needs a TODO - to investigate not 'wrong' - pls just decde

target/rv32imc/riscvOVPsim.ic
simon5656: the target/rv32imc/riscvOVPsim.ic is derived from the riscv-config yaml specs that defines all legal options that a device can have and still meet the riscv specs.

in the riscvOVPsim.ic files there are several important statements that define the extensions. these are all documented in the riscvOVPsim docs, and also if you use , eg
   cd riscv-ovpsim/bin/Linux64
    riscvOVPsim.exe --showoverrides.

--variant RVB32I   # says this is a base rv32i - ie no U no S etc...  - the simplest.
--override riscvOVPsim/cpu/add_Extensions=MC # says add on M and C extensions
--override riscvOVPsim/cpu/misa_MXL=1 # and just confirm 32bits

so yes, to answer your question, Does this include RV32 M,C extension? it does.

thanks
Simon

scripts/ovpsim_log_to_trace_csv.py
taoliug: OK, got it. I have fixed a few issues over the last couple of days. You can try to simulate with top of the tree to see if the issue is gone.

PR #251, opened 2019-10-23
Updates for python subprocess and dsim (systemverilog)

eroom1966:
Creation of subprocess on some machines does not use /bin/bash, define the default shell to be /bin/bash
disabled an unsupported construct for dsim

### DIFF
--- a/src/riscv_directed_instr_lib.sv
+++ b/src/riscv_directed_instr_lib.sv
@@ -206,9 +206,15 @@ class riscv_jal_instr extends riscv_rand_instr_stream;
     order.shuffle();
     setup_allowed_instr(1, 1);
     jal = {JAL};
-    if (!cfg.disable_compressed_instr) begin
-      jal = (XLEN == 32) ? {jal, C_J, C_JAL} : {jal, C_J};
-    end
+	`ifdef DSIM
+	// if (!cfg.disable_compressed_instr) begin
+	//   jal = (XLEN == 32) ? {jal, C_J, C_JAL} : {jal, C_J};
+	// end
+	`else
+		if (!cfg.disable_compressed_instr) begin
+			jal = (XLEN == 32) ? {jal, C_J, C_JAL} : {jal, C_J};
+		end
+	`endif
     // First instruction
     jump_start = riscv_instr_base::type_id::create("jump_start");
     `DV_CHECK_RANDOMIZE_WITH_FATAL(jump_start,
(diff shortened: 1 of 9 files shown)


src/riscv_directed_instr_lib.sv
```diff
@@ -206,9 +206,15 @@ class riscv_jal_instr extends riscv_rand_instr_stream;
     order.shuffle();
     setup_allowed_instr(1, 1);
     jal = {JAL};
-    if (!cfg.disable_compressed_instr) begin
-      jal = (XLEN == 32) ? {jal, C_J, C_JAL} : {jal, C_J};
-    end
+	`ifdef DSIM
```
taoliug: Can you explain what's the issue with this code? We can change the coding style to make it work for dsim

eroom1966:
At the moment it fails to compile
We will get this fixed, but in the meantime we want to fully test with dsim, and with our partners, at the moment the infrastructure is dependant on the google git repo, and changing the infrastructure would be a large undetaking (especially before dvconn next week)
We will return to this and fix properly

taoliug:
I fixed it by changing the coding style. 
BTW, I found ovpsim simulation is broken with default setting (rv64gc). Can you add a ovpsim configure file for this?

eroom1966:
@taoliug 

> I fixed it by changing the coding style.

Have you committed this fix ?
It still fails for me if I download the Master branch

taoliug:
OVPSIM sim failure is now fixed, please refer to this issue
https://github.com/google/riscv-dv/issues/259

The main problem is the cfg make mepc RO and fixed to some old ibex MPEC address. ibex design has been changed to make MEPC R/W. BTW, if you want to have some ibex specific setting, please submit the change to ibex repo, not riscv-dv repo.

taoliug:
Close this PR for now, please file a bug if dsim doesn't work for you

eroom1966:
@taoliug 
Hi Tao, it looks like you did not merge the change to lib.py ?

Without this update, the type of shell which runs the python subcommand is undefined (probably defaults to system/users default shell?), on some machines I run, this breaks the redirect of the logfile for running ovpsim or spike, so we end up with an empty log - this is probably due to the logfile being captured using '&>' - this works for bash, not sure about other shells

>     ps = subprocess.Popen("exec " + cmd,
>                           executable='/bin/bash',

I am guessing the issue is that the default shell is '/bin/sh', on some systems this links to '/bin/bash', but on debian for example, this links to '/bin/dash'

Redirect in /bin/bash is '&>' but in /bin/dash, it appears to be '>&'
In which case, the capturing of the log needs to be modified, or the subprocess shell type needs to be defined

Thx
Lee

PR #258, opened 2019-10-24
Update illegal system instr generation

udinator:
Added a new type of illegal instruction: `kIllegalSystemInstr`.
This is meant to stress invalid `SYSTEM`-type instructions that present both a valid `opcode` and `funct3` field to the processor's decoding logic.

### DIFF
--- a/src/riscv_illegal_instr.sv
+++ b/src/riscv_illegal_instr.sv
@@ -29,7 +29,8 @@ class riscv_illegal_instr extends uvm_object;
     kIllegalFunc3,
     kIllegalFunc7,
     kReservedCompressedInstr,
-    kHintInstr
+    kHintInstr,
+    kIllegalSystemInstr
   } illegal_instr_type_e;
 
   // Default legal opcode for RV32I instructions
@@ -65,15 +66,17 @@ class riscv_illegal_instr extends uvm_object;
   rand bit [1:0]             c_op;
   rand bit [2:0]             c_msb;
   riscv_instr_gen_config     cfg;
+  privileged_reg_t           csrs[$];
 
   constraint exception_dist_c {
     exception dist {
-      kIllegalOpcode           := 4,
+      kIllegalOpcode           := 3,
       kIllegalCompressedOpcode := 1,
       kIllegalFunc3            := 1,
       kIllegalFunc7            := 1,
       kReservedCompressedInstr := 1,
-      kHintInstr               := 3
+      kHintInstr               := 3,
+      kIllegalSystemInstr      := 3
     };
   }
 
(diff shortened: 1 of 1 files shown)


src/riscv_illegal_instr.sv
```diff
@@ -97,6 +99,32 @@ class riscv_illegal_instr extends uvm_object;
...
+      if (func3 == 3'b000) {
+        // Constrain RS1 and RD to be non-zero
+        instr_bin[19:15] != 0;
+        instr_bin[11:7] != 0;
+        // Valid SYSTEM instructions considered by this
+        // Constrain the upper 12 bits to be invalid
+        !(instr_bin[31:20] inside {12'h0, 12'h1, 12'h2, 12'h102, 12'h302, 12'h7b2, 12'h105});
+      } else {
+        // Invalid CSR instructions
+        !(instr_bin[31:20] inside {implemented_csr});
```
taoliug: Will this cause problem since spike/ovpsim may have a different implementation?

udinator: No, it shouldn't be an issue, as the upper 12 bits of the instruction should be the csr_address as mandated by the privileged specification, and given that with this constraint, the rest of the csr instruction is completely valid.
Currently, this change only works with Spike, ovpsim appears to be trapping back to the `start` label at `80000080` every time an illegal instruction is decoded, this might be some configuration issue, though.

taoliug: ovpsim issue has been fixed, it's a config file issue

src/riscv_illegal_instr.sv
```diff
@@ -97,6 +99,32 @@ class riscv_illegal_instr extends uvm_object;
...
+        // Further constrain the csr address to be within the legal range of CSR addresses
```
taoliug: Should the address be constrained to a illegal address? My understanding is that you want to get an address which is invalid for any processor/ISS?

udinator: That was what I considered at first, but I feel it might be better to constrain it to be an unimplemented CSR , since a core should still consider this to be invalid, and it would provide a slightly bigger randomization space.

src/riscv_illegal_instr.sv
udinator: After pulling latest changes, I'm seeing an error that riscvOVPsim.ic is not being found.

src/riscv_illegal_instr.sv
udinator: Resolving for now - changing this to be just an illegal csr address, instead of just unimplemented.

src/riscv_illegal_instr.sv
```diff
@@ -97,6 +100,32 @@ class riscv_illegal_instr extends uvm_object;
...
+      } else {
+        // Invalid CSR instructions
+        !(instr_bin[31:20] inside {csrs});
+        // Further constrain the csr address to be within the legal range of CSR addresses
+        // U-mode CSR address ranges
+        instr_bin[31:20] inside {[12'h0:12'hFF], [12'h400:12'h4FF], [12'h800:12'h8FF], [12'hC00:12'hCFF]};
+        // S-mode CSR address ranges
+        instr_bin[31:20] inside {[12'h100:12'h1FF], [12'h500:12'h5FF], [12'h900:12'h9FF], [12'hD00:12'hDFF]};
+        // M-mode CSR address ranges
+        instr_bin[31:20] inside {[12'h300:12'h3FF], [12'h700:12'h7FF], [12'hB00:12'hBFF], [12'hF00:12'hFFF]};
```
taoliug: Do you still need above 3 constraints?

udinator: Probably not, they have been removed.

taoliug:
Is there any noticeable randomization performance impact of this change?

udinator:
> Is there any noticeable randomization performance impact of this change?

Not really, it takes around 4-8 seconds to generate an illegal_instr_test with this constraint added.

PR #269, opened 2019-10-25
Enable mie.mtie for timer interrupts

udinator:


udinator:
> Does the test pass with spike cosim?

It still doesn't, but it passes with OVPsim as of right now.

taoliug:
> > Does the test pass with spike cosim?
> 
> It still doesn't, but it passes with OVPsim as of right now.

Can you create a separate knob for timer interrupt? Default turn it off and you can enable it in ibex regression.

udinator:
> > > Does the test pass with spike cosim?
> > 
> > 
> > It still doesn't, but it passes with OVPsim as of right now.
> 
> Can you create a separate knob for timer interrupt? Default turn it off and you can enable it in ibex regression.

Done.

PR #273, opened 2019-10-29
Add dummy writes to status and ie CSRs

udinator:


src/riscv_asm_program_gen.sv
```diff
@@ -384,6 +385,37 @@ class riscv_asm_program_gen extends uvm_object;
     end
   endfunction
 
+  // Generate some dummy writes to xSTATUS/xIE at the beginning of the test to check
+  // repeated writes to these CSRs.
+  virtual function void gen_dummy_csr_write();
+    string instr[$];
+    case (cfg.init_privileged_mode)
+      MACHINE_MODE: begin
+        instr.push_back($sformatf("csrr x%0d, 0x%0x", cfg.gpr[0], MSTATUS));
```
taoliug: Will this cause mismatch between ibex and spike?

udinator: No, this shouldn't result in any mismatches

taoliug: Can you run a test in riscv-dv with --iss spike,ovpsim to make sure it's not causing any mismatch between spike and ovpsim?

udinator: Yep, I've ran these tests with both simulators, everything is fine and passing correctly.

PR #274, opened 2019-10-31
fixed trace handling issues

eroom1966:
changes to fix some trace handling problems
changed cfg_path to variant for selecting correct configuration file to riscvOVPsim

### DIFF
--- a/yaml/iss.yaml
+++ b/yaml/iss.yaml
@@ -21,7 +21,7 @@
   path_var: OVPSIM_PATH
   cmd: >
     <path_var>/riscvOVPsim.exe
-    --controlfile <cfg_path>/riscvOVPsim.ic
+    --controlfile target/<variant>/riscvOVPsim.ic
     --objfilenoentry <elf>
     --override riscvOVPsim/cpu/PMP_registers=0
     --override riscvOVPsim/cpu/simulateexceptions=T
--- a/run.py
+++ b/run.py
@@ -93,10 +93,7 @@ def parse_iss_yaml(iss, iss_yaml, isa, setting_dir):
       logging.info("Found matching ISS: %s" % entry['iss'])
       cmd = entry['cmd'].rstrip()
       cmd = re.sub("\<path_var\>", get_env_var(entry['path_var']), cmd)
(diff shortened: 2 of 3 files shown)


yaml/iss.yaml
```diff
@@ -21,7 +21,7 @@
   path_var: OVPSIM_PATH
   cmd: >
     <path_var>/riscvOVPsim.exe
-    --controlfile <cfg_path>/riscvOVPsim.ic
+    --controlfile target/<variant>/riscvOVPsim.ic
```
taoliug: This won't work if we run the script without "--target"

eroom1966: @taoliug 
There needs to be a way to set all the different tool configs that are needed for each target.
the work on riscv-config (https://github.com/riscv/riscv-config) means there is a dut.yaml file that defines the device under test - and this can be used to derive what is needed for the riscv-dv system and all its constituent parts.
Even without the definition of everything from yaml there needs to be a clean way in riscv-dv to set up the different tools and subsystems for the different targets/duts

For example, today for a target there are the following items need setting

- core_setting dir for the riscv_core_setting.sv  to define the dut for the generator
- testlist.yaml to define the tests for the dut
- isa used set variant of spike
- mabi used by the gcc
- riscvOVPsim.ic ovpsim control file to define the dut very specifically

so I think there needs to be some gardening done on riscv-dv to tidy things up
ideally there should be a dut.yaml that you create all the above from
and the use of the command line options for core_setting_dir, isa, mabi etc should be deprecated

for now the --target and target dir should be used and all runs need the --target option - or a default target dir can be used and all files and dut config data should be in the target dir or their location defined in the dut.yaml and a --target_yaml option points to it.

until you do the above refactoring ovpsim needs to be like spike and make use of the --isa to select the target etc. so please allow our usage of variant in iss.yaml and fix run.py to provide necessary default so [as you require] the optional use of --target works

thanks

taoliug: I agree the setup right now is a bit complicated, I will merge this PR, and have a subsequent PR for the default target.

PR #282, opened 2019-11-04
Re-enable custom OVPsim configuration files

udinator:


simon5656:
Hi
Not sure what is going on here as this is just reverting to break all our tests (again!)

with run.py and yaml/iss.yaml using variant for ovpsim we use it from our makefiles with:

TARGET_ISA ?= rv32i
TARGET_CORE_SETTING_DIR ?= $(TOPDIR)/riscv-dv/target
python3 ./run.py -o ./outdir \
    --isa $(TARGET_ISA) \
    --core_setting_dir $(TARGET_CORE_SETTING_DIR)/$(TARGET_ISA) \
    --testlist $(TARGET_CORE_SETTING_DIR)/$(TARGET_ISA)/testlist.yaml
    ...

so if you revert this change and use the above options it should work for you.
thanks
Simon

To: google/riscv-dv <riscv-dv@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Subject: [google/riscv-dv] Re-enable custom OVPsim configuration files (#282)

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/google/riscv-dv/pull/282

Commit Summary

  *   Re-enable custom OVPsim configuration files

File Changes

  *   M run.py<https://github.com/google/riscv-dv/pull/282/files#diff-0> (5)
  *   M yaml/iss.yaml<https://github.com/google/riscv-dv/pull/282/files#diff-1> (2)

Patch Links:

  *   https://github.com/google/riscv-dv/pull/282.patch
  *   https://github.com/google/riscv-dv/pull/282.diff

taoliug:
Hi Simon,

Let me check if Ican make a PR to make the recent change backward compatible (which is the intention). Generally we are planning to deprecate the core_setting_dir option, and replace it with "--target" or "--custom_target".  Seems like you are using pre-defined targets, I think you can simply use "--target ${TARGET_ISA}" and  remove the --core_setting_dir and --testlist option

Regards,
Tao

simon5656:
Yes Tao - I think it would be best if it was just --target and that specifies a dir under /target eg /target/<dut_name> where all target specific files can be found, and then if --target is not given then there is a /target/<name> which is where the default settings are - (so there are not defaults in the .py's just in the default --target files etc...
thanks
Simon
[also - it would be good if there was simple set of regression tests for different options - as part of the repo - we will think on this.]
thanks
Simon

________________________________

taoliug:
For the RTL processor specific target, right now it's implemented as a custom target. Take ibex for example, you can find the custom target directory here. 
https://github.com/lowRISC/ibex/blob/master/dv/uvm/Makefile#L67
You can see the custom changes are more than just testlist, core_setting, ovpsim config etc:
https://github.com/lowRISC/ibex/tree/master/dv/uvm/riscv_dv_extension
That's why we prefer to put it under ibex repo rather than riscv-dv repo.  Besides, the processor may involve over time, I feel it's better to let the processor repo owner to maintain the custom target.

simon5656:
Yes

all device specific files needs to be able to be selected to be outside of riscv-dv

we had it so that we had a side dir with all the needed device specific files in it and used the commands line options to select them - this was for iss, simulator, testlist, etc...

there should be files with default versions of all the options - but every one needs to be able to be over-ridden in the command line - and their location needs to be able to be given

I don't believe there should complex default choices being made in run.py and cov.py - it needs to be done by user from command line settings.

I believe there is re-factoring needed for all the command options as the cov.py is not compatible with the run.py options.

the current repo state is broken in two ways
1) we cant set testlist for the coverage
2) --isa and --mabi cant be passed in from cov to run...

needs refactoring.

I will come up with a  spec of what options I believe are needed.

any user should be able to do what they need for a device and tool options without having to touch a line of code in the riscv-dv repo.

thanks
Simon


________________________________

PR #285, opened 2019-11-06
start of vector intruction processing for coverage

simond-imperas:
This is the start of adding coverage

### DIFF
--- a/cov.py
+++ b/cov.py
@@ -39,7 +39,7 @@ def collect_cov(log_dir, out, iss, testlist, batch_size, lsf_cmd, steps, \
     log_dir          : ISS log directory
     out              : Output directory
     iss              : Instruction set simulator
-    test_list        : Testlist of the coverage test
+    testlist        : Testlist of the coverage test
     batch_size       : Number of trace CSV to process per test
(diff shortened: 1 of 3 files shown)


cov.py
```diff
@@ -268,6 +268,8 @@ def main():
   else:
     args.testlist = args.custom_target + "/testlist.yaml"
 
+  # args.testlist = cwd + "/yaml/cov_testlist.yaml" ## needed if need to force
```
taoliug: The testlist for coverage regression should be fixed for all targets

simond-imperas:
Tao

yes but there is something wrong with all the if else - -as without out
this line

args.testlist = cwd + "/yaml/cov_testlist.yaml" ## needed if need to force

to force it - we could not stop it trying target rv32i - and we are doing a
rv32imc etc

just leave the comment in for now - and when you refactor it not nbe needed
and will all be sorted.

just means we need this for the time being.

thanks
Simon

simon5656:
Tao - can you delete this pull request - I will submit again this time with a corrected committer to try that - thanks Simon

PR #291, opened 2019-11-09
Enable using core trace logs for coverage collection

udinator:


taoliug:
Can you resolve your PR with the upstream changes?

udinator:
> Can you resolve your PR with the upstream changes?

Done.

cov.py
```diff
@@ -68,22 +75,33 @@ def collect_cov(log_dir, out, iss, testlist, batch_size, lsf_cmd, steps, \
...
-      logging.info("Process %0s log[%0d/%0d] : %s" % (iss, i+1, len(log_list), log))
-      if iss == "spike":
-        process_spike_sim_log(log, csv, 1)
-      elif iss == "ovpsim":
-        process_ovpsim_sim_log(log, csv, 1, stop_on_first_error)
+      # If a core target is defined, prioritize over ISS
+      if core:
+        logging.info("Process %0s log[%0d/%0d] : %s" % (core, i+1, len(log_list), log))
+        if core == "ibex":
+          ibex_log_cmd = ("python3 %s/ibex_log_to_trace_csv.py --log %s --csv %s --full_trace 1" %
```
taoliug: Can we have a way to extend core specific behavior in ibex repo? I think it's better to not include core specific code in riscv-dv.

udinator: Something that might be easier actually, is to default enable the full trace parsing of the ibex logs, so that calling the parsing scripts again is not necessary, what do you think?

taoliug: yes, default enable full_trace sounds good to me.

udinator: Done, see https://github.com/lowRISC/ibex/pull/453

cov.py
```diff
@@ -31,33 +31,40 @@
...
+    target              : Predefined target
     stop_on_first_error : will end run on first error detected
   """
   cwd = os.path.dirname(os.path.realpath(__file__))
   log_list = []
   csv_list = []
-  trace_log = ("%s/%s_trace_log" % (out, iss))
   logging.info("Processing trace log under %s" % log_dir)
-  run_cmd("find %s -name \"*.log\" | sort > %s" % (log_dir, trace_log))
+  if core:
```
taoliug: Do you need this?

cov.py
```diff
@@ -215,6 +228,12 @@ def setup_parser():
                       help="Number of CSV to process per run")
   parser.add_argument("-s", "--steps", type=str, default="all",
                       help="Run steps: csv,cov", dest="steps")
+  parser.add_argument("--core", type=str, default="",
+                      help="Name of target core")
+  parser.add_argument("--isa", type=str, default="",
```
taoliug: Why --isa and --mabi is needed? I believe we are not compiling or simulating ISS for the coverage flow?

cov.py
udinator: I use this to properly name the output file `trace_log`, as well as filter out some extra files from the `find` command, I specify that logfiles should have `trace` somewhere in their name, as this directory also contains vcs simulation log files.

cov.py
udinator: That's because I specific custom_target, and run.py requres mabi/isa to be set if custom_target is defined.

taoliug: I think we should remove this requirement in the run.py if the steps don't include gcc_compile and iss_sim

cov.py
taoliug: I think you don't need log anymore, just need a list of CSV files?

cov.py
udinator: Agreed, I'm making this change right now.

cov.py
udinator: yes, you're right, I missed getting rid of this code after enabling full_trace by default. will fix this now

cov.py
```diff
@@ -31,51 +31,65 @@
...
+    If functional coverage is being collected from an RTL core implementation,
+    the flow assumes that the core's trace logs have already been converted to
+    CSV files by the post_compare step of the flow.
+    """
+    trace_log = ("%s/%s_trace_log" % (out, core))
+    run_cmd("find %s -name \"*.csv\" | sort > %s" % (log_dir, trace_log))
+  else:
+    trace_log = ("%s/%s_trace_log" % (out, iss))
+    run_cmd("find %s -name \"*.log\" | sort > %s" % (log_dir, trace_log))
+    with open(trace_log) as f:
```
taoliug: I believe you need to remove the indent here?

udinator: Yes, I've removed it locally. I found a bug with how I'm using assign_operand(), so I haven't updated this PR yet

PR #293, opened 2019-11-09
added option to cov.py to stop on first error

simond-imperas:
currently only for quest32 and covpy
other simulators need adding
future is this could be aded to lib.py and used in rub.py too
currently only opertional in verbose


cov.py
```diff
@@ -31,8 +31,27 @@
 
 LOGGER = logging.getLogger()
 
+def check_simulator_return(output, simulator, stop_on_first_error):
```
taoliug: I think this should be moved to lib.py

simond-imperas: ok will do

cov.py
```diff
@@ -231,8 +235,11 @@ def setup_parser():
                       help="Directory name of the custom target")
   parser.add_argument("-cs", "--core_setting_dir", type=str, default="",
                       help="Path for the riscv_core_setting.sv")
+  parser.add_argument("--stop_on_first_error", dest="stop_on_first_error", action="store_true",
```
taoliug: Fix alignment here.

scripts/lib.py
```diff
@@ -172,3 +172,20 @@ def process_regression_list(testlist, test, iterations, matched_list):
...
+    TODO add other simulators
+    """
+
+    if not stop_on_first_error: return
+
+    if "questa32" == simulator:
+        for line in output.splitlines():
+            if "Errors: " in line:
+                if not "Errors: 0" in line:
+                    logging.fatal ("check_simulator_return (%s): TERMINATING as got errors: [%s]" %
```
taoliug: Fix line length here

scripts/lib.py
```diff
@@ -172,3 +172,20 @@ def process_regression_list(testlist, test, iterations, matched_list):
...
+def check_simulator_return(output, simulator, stop_on_first_error):
+    """
+    tests simulator output for errors and terminates run if found
+    ONLY works when verbose is on (as output not returned otherwise)
+    TODO add other simulators
+    """
+
+    if not stop_on_first_error: return
+
+    if "questa32" == simulator:
```
taoliug: Should this be "questa"?
https://github.com/google/riscv-dv/blob/master/yaml/simulator.yaml#L47

simond-imperas: the name in our simulator.yaml is questa32 (which runs remotely on 32bit machine)  thus not the questa in your simulator.yaml - so it is the only one that I have tested to catch errors - so for us it needs to be questa32 (or coul I use an alias type of mechanism in my simulator.yaml

simond-imperas: I have made it 
> if "questa" in simulator:

 but have only tested on our 32 bit hosted version.

PR #304, opened 2019-11-14
Basic U-mode support

udinator:


src/riscv_asm_program_gen.sv
```diff
@@ -542,6 +541,9 @@ class riscv_asm_program_gen extends uvm_object;
             gen_signature_handshake(.instr(csr_handshake), .signature_type(WRITE_CSR), .csr(UIE));
           end
         endcase
+        // Write M-mode CSRs to testbench by default, as these should be implemented
```
taoliug: Can you access M mode register in U mode trap?

udinator: The `init_[m/s/u]mode` sections are always executed in M mode directly out of reset, the intended mode is switched into after the xRET instruction at the end of this section

PR #316, opened 2019-11-19
Add directed stream to access higher privilege CSRs

udinator:
- All invalid (e.g. higher privilege level) CSRs are added to a queue in `cfg` on generator initialization, making constraints much more simplified

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -389,34 +389,36 @@ class riscv_asm_program_gen extends uvm_object;
   // repeated writes to these CSRs.
   virtual function void gen_dummy_csr_write();
     string instr[$];
-    case (cfg.init_privileged_mode)
-      MACHINE_MODE: begin
-        instr.push_back($sformatf("csrr x%0d, 0x%0x", cfg.gpr[0], MSTATUS));
-        instr.push_back($sformatf("csrr x%0d, 0x%0x", cfg.gpr[1], MIE));
-        instr.push_back($sformatf("csrw 0x%0x, x%0d", MSTATUS, cfg.gpr[0]));
-        instr.push_back($sformatf("csrw 0x%0x, x%0d", MIE, cfg.gpr[1]));
-      end
-      SUPERVISOR_MODE: begin
-        instr.push_back($sformatf("csrr x%0d, 0x%0x", cfg.gpr[0], SSTATUS));
-        instr.push_back($sformatf("csrr x%0d, 0x%0x", cfg.gpr[1], SIE));
-        instr.push_back($sformatf("csrw 0x%0x, x%0d", SSTATUS, cfg.gpr[0]));
-        instr.push_back($sformatf("csrw 0x%0x, x%0d", SIE, cfg.gpr[1]));
-      end
-      USER_MODE: begin
-        if (!riscv_instr_pkg::support_umode_trap) begin
-          return;
+    if (cfg.enable_dummy_csr_write) begin
+      case (cfg.init_privileged_mode)
+        MACHINE_MODE: begin
+          instr.push_back($sformatf("csrr x%0d, 0x%0x", cfg.gpr[0], MSTATUS));
+          instr.push_back($sformatf("csrr x%0d, 0x%0x", cfg.gpr[1], MIE));
+          instr.push_back($sformatf("csrw 0x%0x, x%0d", MSTATUS, cfg.gpr[0]));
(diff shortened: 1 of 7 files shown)


src/riscv_asm_program_gen.sv
```diff
@@ -385,38 +386,74 @@ class riscv_asm_program_gen extends uvm_object;
     end
   endfunction
 
+  virtual function void gen_access_higher_csr_level();
```
taoliug: It's OK to do it like this for now, but I'd expect to enable this type of illegal access randomly in the main program later.

src/riscv_instr_gen_config.sv
```diff
@@ -403,6 +408,8 @@ class riscv_instr_gen_config extends uvm_object;
     get_bool_arg_value("+no_load_store=", no_load_store);
     get_bool_arg_value("+no_csr_instr=", no_csr_instr);
     get_bool_arg_value("+enable_illegal_csr_instruction=", enable_illegal_csr_instruction);
+    get_bool_arg_value("+enable_access_higher_csr_level=", enable_access_higher_csr_level);
```
taoliug: Can you add a test to enable this option for the rv64gc target?

src/riscv_asm_program_gen.sv
```diff
@@ -385,38 +386,74 @@ class riscv_asm_program_gen extends uvm_object;
...
+    string instr[$];
+    string valid_level[$];
```
taoliug: I'd prefer to rename it to invalid_level as this is actually illegal access. Please also add a comment about what this task trying to do.

src/riscv_asm_program_gen.sv
udinator: Good point, let me actually just make that feature for this PR, should just be some modifications to the instruction randomization constraints.

taoliug: I think we need to avoid generating a lot of illegal access which could take a very long time to run, or you could limit the number of instruction for that test.

udinator: Would just keeping the normal randomization be okay for this case, since we would explicitly enable this via a config knob?

taoliug: yes, sgtm

udinator: Updated to enable this feature in the instruction randomization code.

src/riscv_asm_program_gen.sv
udinator: This function is now deleted, to match the feature changes to allow this invalid csr access being randomized on an instruction level, instead of a directed sequence

src/riscv_instr_gen_config.sv

src/riscv_rand_instr.sv
```diff
@@ -95,19 +95,38 @@ class riscv_rand_instr extends riscv_instr_base;
...
     } else {
       if (cfg.enable_illegal_csr_instruction) {
         !(csr inside {implemented_csr});
+      } else if (cfg.enable_access_invalid_csr_level) {
+        if (cfg.init_privileged_mode == MACHINE_MODE) {
+          // Invalid privilege levels from M-mode are Debug mode
+          csr inside {[TSELECT:DSCRATCH1]};
+        } else if (cfg.init_privileged_mode == SUPERVISOR_MODE) {
+          // Invalid privilege levels from S-mode are M-mode and Debug mode
+          csr inside {[MVENDORID:MHARTID], [MSTATUS:PMPADDR15],
```
taoliug: Why do you need to list them? can you still use the original approach to get a list of illegal registers based on the privileged mode?

udinator: I had considered something like that, essentially compiling a list of "m"/"s"/"u"/"d" CSRs, but I wanted to be able to include the performance and trigger CSRs, so I went with the manual listing. I could potentially exclude the performance/trigger CSRs from this randomization, which would allow me to use the first method

udinator: This functionality has been moved to `riscv_instr_gen_config.sv`.

target/rv64gc/testlist.yaml
```diff
@@ -53,6 +53,18 @@
     +num_of_sub_program=5
   rtl_test: core_base_test
 
+- test: riscv_invalid_csr_test
```
taoliug: missing iteration count?

udinator: You're right, I've added the base iteration count of 2.

PR #317, opened 2019-11-20
Added command line control of coverage and added hooks for vector coverage development

simond-imperas:
Added options to pass parameters into cov.py which sets `defines and +args for controlling the inclusion or not of coverpoints and sampling etc. - when this is merged in and nothing changed - then I will add some documentation on how to set things.

Also added into the coverage .sv files a lot of includes that are defined out if vectors are not enabled - this allows us to continue to merge in changes made in this repo into our local code without issue - and when we have our code refined and stable we can merge it in - this approach means there is no issues for others as our code is edited and churns - allows us to develop with no repurcussions on the rest of riscv-dv users.

I also fixed issue with jalr and j., c. instructions in the ovpsim log to csv - where ovpsim has hex numbers in its trace that do not have leading 0x - and so these do need adding in... - if this breaks something for you - pls send me the .log with them in - and I can sort.
thanks


taoliug:
One general comment, please make future PR smaller and single purpose. This will make review easier and faster.

src/riscv_instr_base.sv
```diff
@@ -449,6 +451,8 @@ class riscv_instr_base extends uvm_object;
   // Supervisor Instructions
   `add_instr(SFENCE_VMA, R_FORMAT,SYNCH,RV32I)
 
+  `VECTOR_INCLUDE("riscv_instr_base_inc_5.sv")
```
taoliug: Can you give more meaningful name to these files?

src/riscv_instr_pkg.sv
```diff
@@ -86,10 +86,12 @@ package riscv_instr_pkg;
...
     RV64C,
     RV128I,
-    RV128C
+    RV128C,
+    RV32V,
+    RV64V
   } riscv_instr_group_t;
 
-  typedef enum {
+typedef enum {
```
taoliug: Alignment is off here

test/riscv_instr_cov_test.sv
```diff
@@ -36,8 +37,6 @@ class riscv_instr_cov_test extends uvm_test;
       i++;
     end
     cfg = riscv_instr_gen_config::type_id::create("cfg");
-    // disable_compressed_instr is not relevant to coverage test
-    cfg.disable_compressed_instr = 0;
```
taoliug: What's the purpose of this change?

test/riscv_instr_cov_test.sv
```diff
@@ -46,12 +45,8 @@ class riscv_instr_cov_test extends uvm_test;
     instr_cg = new(cfg);
     `uvm_info(`gfn, $sformatf("%0d CSV trace files to be processed", trace_csv.size()), UVM_LOW)
     foreach (trace_csv[i]) begin
-      bit expect_illegal_instr;
```
taoliug: Here and below, looks like the change is not intended?

scripts/ovpsim_log_to_trace_csv.py
```diff
@@ -55,6 +55,11 @@ def convert_mode(pri, line):
         "a0","a1","a2","a3","a4","a5","a6","a7",
         "s2","s3","s4","s5","s6","s7","s8","s9","s10","s11",
         "t3","t4","t5","t6"]
+FREGS = ["ft0","ft1","ft2","ft3","ft4","ft5","ft6","ft7","fs0",
+        "fs1",
```
taoliug: One more space here. Besides, make the line length more uniform.

test/riscv_instr_cov_test.sv
```diff
@@ -133,14 +132,19 @@ class riscv_instr_cov_test extends uvm_test;
...
+    unexpected_illegal_instr_cnt++;
+    fatal($sformatf(
+        "FATAL: sample(%0s) is ILLEGAL instruction (for the included ISAs)",
+            trace["instr"]));
     return 1'b0;
   endfunction
 
   virtual function void assign_trace_info_to_instr(riscv_instr_cov_item instr);
     riscv_reg_t gpr;
+   `VECTOR_INCLUDE("riscv_instr_cov_test_inc_2.sv")
```
taoliug: I don't see these files included in this PR, is it intended?

test/riscv_instr_cov_test.sv
```diff
@@ -108,12 +95,24 @@ class riscv_instr_cov_test extends uvm_test;
...
+      `uvm_info(`gfn, "TEST PASSED", UVM_NONE);
     end
   endtask
 
   virtual function void post_process_trace();
   endfunction
 
+  function void fatal (string str);
+    `uvm_info(`gfn, str, UVM_NONE);
+    `ifdef STOP_ON_FIRST_ERROR
```
taoliug: Prefer runtime option rather than compile time option.

src/riscv_instr_base.sv
simond-imperas: hard I think as they are just fragments and they are positional and their existence is temporary while development takes place - I guess I could say filename_approx_line - but  they say file and that they are an include - they are just slots in code to add fragments - I cold say 'function_name_near_top, or class_def_near_bottom... thats why I just went for filename_inc_num... what would you suggest if you feel it is needed?

test/riscv_instr_cov_test.sv
simond-imperas: I don't know - I guess this was removed/added in a previous commit - are you ok for me to just remove it?

src/riscv_instr_pkg.sv
simond-imperas: ok

test/riscv_instr_cov_test.sv
simond-imperas: or should it be there?

test/riscv_instr_cov_test.sv
simond-imperas: yes - not sure how this happens - I don't recall editing or seeing this. - I have reinstated

scripts/ovpsim_log_to_trace_csv.py
simond-imperas: ok

test/riscv_instr_cov_test.sv
simond-imperas: yes it is intended that they do not need to exist (I have them in a user extension in a different place outside this repo as I work on them) - and there is a `define that needs to be set for these files to be included (hence inside the `VECTOR_INCLUDE macro). 
so they do not need to exist, and so there is no need to include empty files.

test/riscv_instr_cov_test.sv
simond-imperas: ok - you mean a +stop_on_first_error?

taoliug: yes

test/riscv_instr_cov_test.sv
taoliug: Yes, please leave it there

simond-imperas:
OK - so I think I resolved all your items. thanks Simon

test/riscv_instr_cov_test.sv
```diff
@@ -250,23 +265,4 @@ class riscv_instr_cov_test extends uvm_test;
     end
   endfunction
 
-  function void report_phase(uvm_phase phase);
```
taoliug: Please revert this change as well

test/riscv_instr_cov_test.sv
```diff
@@ -81,17 +82,9 @@ class riscv_instr_cov_test extends uvm_test;
...
-                            entry_cnt, trace["str"]), UVM_FULL)
             if (!sample()) begin
-              if (expect_illegal_instr) begin
-                `uvm_info(`gfn, $sformatf("Found illegal instr: %0s [%0s]",
-                                trace["instr"], line), UVM_HIGH)
-              end else begin
-                unexpected_illegal_instr_cnt += 1;
-                `uvm_error(`gfn, $sformatf("Found illegal instr: %0s [%0s]",
-                                 trace["instr"], line))
-              end
```
taoliug: Why making this change?

scripts/riscv_trace_csv.py
```diff
@@ -117,12 +124,18 @@ def write_trace_entry(self, entry):
                               'vs2_val'    : entry.vs2_val,
                               'vs3'        : entry.vs3,
                               'vs3_val'    : entry.vs3_val,
-                              'vtype_e'    : entry.vtype_e,
-                              'vtype_m'    : entry.vtype_m,
-                              'vtype_d'    : entry.vtype_d,
+                              'vtype_vsew'    : entry.vtype_vsew,
```
taoliug: Alignment, here and below

src/riscv_asm_program_gen.sv
```diff
@@ -340,7 +340,7 @@ class riscv_asm_program_gen extends uvm_object;
       init_floating_point_gpr();
     end
     core_is_initialized();
-    gen_dummy_csr_write();
+    gen_dummy_csr_write(); // comment out if not want to read incorrect values from csr (spike problem)
```
taoliug: Please remove this comment, we will add a way to disable xStatus read in this task

src/riscv_instr_base.sv
taoliug: I'd prefer to name them by the content of the file, for example:
riscv_vector_instr_enum.sv
riscv_vector_instr_constraint.sv
...

test/riscv_instr_cov_test.sv
simond-imperas: surely we  want to know that we have found an instruction that the coverage system does not understand or handle - so it should not be silent - I have made it a n info with a UVM_HIGH so it not that visible...

src/riscv_asm_program_gen.sv
simond-imperas: ok - but pls fix this urgently and put an option in as I cant see how anybody can verify their hardware if you are reading bad resutls from the reference... - and this error fails most of my regressions...

test/riscv_instr_cov_test.sv
taoliug: The intention the original code is report a UVM_ERR if illegal instruction is not expected. Does this match what you are looking for? Looks like your change actually turns the error off.

simond-imperas: I am happy the way I have left it

taoliug: I don't quite get it. If there's unexpected illegal instructions, it should be flagged as an error. Right now it's only gets printed when verbosity is set to high. It's demoting the error, is this the intention of this change?

simond-imperas: I am now confused... I have reverted to what you have in the upstream - as I not sure why this got changed.

taoliug: ok, that's fine, I can fix it in a later PR

src/riscv_instr_pkg.sv
```diff
@@ -352,10 +357,11 @@ package riscv_instr_pkg;
     CL_FORMAT,
     CS_FORMAT,
     CSS_FORMAT,
-    CIW_FORMAT
+    `VECTOR_INCLUDE("riscv_instr_pkg_inc_riscv_instr_format_t.sv")
```
taoliug: riscv_vector_instr_format.sv

src/riscv_instr_pkg.sv
```diff
@@ -371,7 +377,8 @@ package riscv_instr_pkg;
     CHANGELEVEL,
     TRAP,
     INTERRUPT,
-    AMO
+    `VECTOR_INCLUDE("riscv_instr_pkg_inc_riscv_instr_category_t.sv")
```
taoliug: riscv_vector_instr_category.sv

test/riscv_instr_cov_test.sv
```diff
@@ -4,6 +4,7 @@ class riscv_instr_cov_test extends uvm_test;
   typedef uvm_enum_wrapper#(riscv_instr_name_t) instr_enum;
   typedef uvm_enum_wrapper#(riscv_reg_t) gpr_enum;
   typedef uvm_enum_wrapper#(privileged_reg_t) preg_enum;
+  `VECTOR_INCLUDE("riscv_instr_cov_test_inc_typedef.sv")
```
taoliug: here are below, I don't think you really need any include in this file. You could have an extended test, for example:

class riscv_vector_instr_cov_test extends riscv_instr_cov_test;

In you extended test, you can add all the custom functionalities. After you have done this, you can add a test entry in the cov.testlist to run with this test. Please let me know your thought about this.
I am hoping to do some refactoring at some time to make the extensions more modular. However, it will be non-trivial effort, I need to have a plan for it.

src/riscv_instr_pkg.sv
```diff
@@ -318,6 +320,7 @@ package riscv_instr_pkg;
     SRET,
     WFI,
     SFENCE_VMA,
+    `VECTOR_INCLUDE("riscv_instr_pkg_inc_riscv_instr_name_t.sv")
     // You can add other instructions here
```
taoliug: riscv_vector_instr_name.sv

src/riscv_instr_pkg.sv
```diff
@@ -331,12 +334,14 @@ package riscv_instr_pkg;
     S2, S3, S4, S5, S6, S7, S8, S9, S10, S11, T3, T4, T5, T6
   } riscv_reg_t;
 
+  `VECTOR_INCLUDE("riscv_instr_pkg_inc_enum.sv")
```
taoliug: riscv_vector_instr_enum.sv

simond-imperas: why? the files names again,  - ok - these are temporary and I believe that the include name needs to be seen to related to the file name that it is included in - it means no error, no thought, and I dont need to dream up new names... - it just works - so rather than me iterate again - pls tell me which exactly names you want me to change and to what - and I will do it - and then we can get this pull request done - thanks for your help here. Simon

test/riscv_instr_cov_test.sv
simond-imperas: Tao,

Yes some of the new include files for vectors do not have a lot of content at this time and yes things can be done several ways like you say extending the test - but I want to check in as few changes as possible into the upstream so I can maintain contact with it and keep my developments up to date. There is much churn in where we put things as we understand the complexities of the vectors and what 'coverage' means for these polymorphic engines - and so I want to keep that churn off your upstream so that we can make rapid progress to something that is good - and then we can merge it in to the upstream without the include files.

Some data: there are over 600 new instructions (including the many variations), 32 new variable width csrs, 5 (at least today) new gprs, 50+ different hardware configurations related to size and functioning ... and the foundation's vector spec changes every week - so we need the most flexible way of keeping up with your upstream while not wasting everybodies time with pull requests that take functionality in and out on a daily basis - and this is while we are working out how the amost 700 coverpoints change... as users define what functional coverage (inc. cross coverage) is important to them...

And also - I am concerned about making too many changes to your code as the repo does inot include regression tests that I can use to confirm I am not breaking your functionality - we use our own rapidly growing regression list here - and don't do a pull request until they all pass.

So please accept this for now - and let us work out what and where we need functionality - in a way that has no impact on you and other developers/users... and we can keep our work in sync with yours and when the vectors and our work stabilizes we can merge the code in to your upstream.

[and with this include mechanism  - the vector includes will not affect you - and we can move them as needed as-and-when for your refactored code when the time comes - no worries]

thanks. Simon

taoliug: Sure, thanks for your patience. I understand the vector extensions is going to touch a lot of code, that's why I am worrying about it could be done without thorough communication between us. I will take this change for now. In the meantime, I am working on some refactoring to make the whole extension flow cleaner. It should be ready in a week. Please hold on your further changes until the refactoring is done, thanksl

src/riscv_instr_pkg.sv
taoliug: Never mind, I will take this for now, and change it as part of my refactoring work.

test/riscv_instr_cov_test.sv
simond-imperas: Thanks - I am hoping there should be only a few more minor changes for now - one will be save cpu state from instruction to instruction - that might be just one more include - I am looking at that now. - Simon

PR #328, opened 2019-11-28
Added two includes and starting variables for adding bitmanip extension

simond-imperas:
In maintaining cpu state from one instruction to the next we need to add some persistent state and its appropriate resets - so added two new include files to do this.
Also, added initial start so can add new instructions for bitmanipulation coverage - again done using the includes so it does not perturb the main functioning of the project repository.
Also have run coverage for RISC-V Foundation's 5 main compliance tests - to allow them to focus on missing tests - so coverage is proving its worth to them.
Thanks. Simon


taoliug:
Simon, thanks for your PR.
Please be aware of the new instruction classes here:
https://github.com/google/riscv-dv/tree/master/src/isa
It allows each ISA extension to be more modular. The randomization performance is better. I'd prefer the vector/Bitmanip extension to be done with these new classes. I am still doing some testing with these new classes. Plan to enable them in a week time.

simond-imperas:
OK thanks - we will look at these new instruction classes and see how to use them - will you merge this current PR or do I need to make any changes?

PR #336, opened 2019-12-04
refactor debug ROM generation

udinator:
- moved debug_rom generation to a new class, extending from riscv_asm_program_gen, and factory override to correct type in the test class
- Change factory override method for users wishing to extend riscv_asm_program_gen class to uvm_set_inst_override to prevent overriding debug class generation as well, updated README with this new method
- prevent umode_intr_handler section from being generated if core does not support u-mode interrupts
- minor changes to ecall_handler section to adjust generated indentation


src/riscv_asm_program_gen.sv
```diff
@@ -1217,208 +1221,15 @@ class riscv_asm_program_gen extends uvm_object;
...
-    string str;
-    str = $sformatf("csrr x%0d, 0x%0x", cfg.scratch_reg, csr);
-    instr.push_back(str);
-    str = $sformatf("addi x%0d, x%0d, 0x%0x", cfg.scratch_reg, cfg.scratch_reg, val);
-    instr.push_back(str);
-    str = $sformatf("csrw 0x%0x, x%0d", csr, cfg.scratch_reg);
-    instr.push_back(str);
+    `uvm_info(`gfn, "Creating debug ROM", UVM_LOW)
+    debug_rom = riscv_asm_program_gen::type_id::create("debug_rom", , {"uvm_test_top", ".", `gfn});
+    debug_rom.cfg = cfg;
```
taoliug: Dose it make sense to use the same config to generate debug ROM? I think debug rom should be much simpler than the main program.

src/riscv_asm_program_gen.sv
```diff
@@ -1217,208 +1221,15 @@ class riscv_asm_program_gen extends uvm_object;
...
-  virtual function void increment_csr(privileged_reg_t csr, int val, ref string instr[$]);
-    string str;
-    str = $sformatf("csrr x%0d, 0x%0x", cfg.scratch_reg, csr);
-    instr.push_back(str);
-    str = $sformatf("addi x%0d, x%0d, 0x%0x", cfg.scratch_reg, cfg.scratch_reg, val);
-    instr.push_back(str);
-    str = $sformatf("csrw 0x%0x, x%0d", csr, cfg.scratch_reg);
-    instr.push_back(str);
+    `uvm_info(`gfn, "Creating debug ROM", UVM_LOW)
+    debug_rom = riscv_asm_program_gen::type_id::create("debug_rom", , {"uvm_test_top", ".", `gfn});
```
taoliug: debug_rom_gen?

udinator: I use riscv_asm_program_gen here and then override it in the riscv_instr_base_test to the debug_rom_gen type

src/riscv_asm_program_gen.sv
udinator: Yes I think it makes sense here, as many of the methods in riscv_asm_program_gen::* utilize various values stored in the config object, and so I use the same config for the debug ROM due to the heavy utilization of these methods - like gen_sub_program(), riscv_instr_sequence::gen_instr(), gen_callstack(), push/pop_gpr_from_kernel_stack(), and so on.

src/riscv_asm_program_gen.sv
taoliug: why not use debug_rom_gen directly?

udinator: This introduces dependency errors, since riscv_debug_rom_gen extends from riscv_asm_program_gen, so directly creating the riscv_debug_rom_gen type in riscv_asm_program_gen.sv causes compilation failure, I'm using the overrides to get around this issue

PR #338, opened 2019-12-07
Fix list access exception thrown when parsing ovpsim illegal instruction

udinator:
also make ovpsim parsing script indentations 2 spaces, to avoid mixing 2-space and 4-space indents in one file

### DIFF
--- a/scripts/ovpsim_log_to_trace_csv.py
+++ b/scripts/ovpsim_log_to_trace_csv.py
@@ -27,34 +27,34 @@ from lib import *
 from riscv_trace_csv import *
 
 try:
-    from ovpsim_log_to_trace_csv_vectors import *
+  from ovpsim_log_to_trace_csv_vectors import *
 except:
-    def assign_operand_vector(a,b,c,d):
-        """ stub version when no vector processing included """
-        logging.info("No OVPsim vector instruction processing included")
-        if stop_on_first_error:
-            sys.exit(-1)
-    def is_an_extension_instruction(instr):
-        if 'v' in instr[0]:
-            return True
-        return False
+  def assign_operand_vector(a,b,c,d):
+    """ stub version when no vector processing included """
+    logging.info("No OVPsim vector instruction processing included")
+    if stop_on_first_error:
+      sys.exit(RET_FATAL)
+  def is_an_extension_instruction(instr):
+    if instr and 'v' in instr[0]:
+      return True
+    return False
 
 
 stop_on_first_error = 0
 
 def fatal (s):
-    """ ensure we end if a problem """
-    logging.fatal("ERROR: "+s)
-    sys.exit(-1)
+  """ ensure we end if a problem """
+  logging.fatal("ERROR: "+s)
+  sys.exit(RET_FATAL)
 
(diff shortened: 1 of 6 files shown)


scripts/ovpsim_log_to_trace_csv.py
```diff
@@ -27,34 +27,35 @@
...
-        if 'v' in instr[0]:
-            return True
-        return False
+  def assign_operand_vector(a,b,c,d):
+    """ stub version when no vector processing included """
+    logging.info("No OVPsim vector instruction processing included")
+    if stop_on_first_error:
+      sys.exit(-1)
+  def is_an_extension_instruction(instr):
+    if instr:
```
danghai: I think it should be `if instr and `v` in instr[0]`

scripts/ovpsim_log_to_trace_csv.py
```diff
@@ -27,34 +27,35 @@
...
-    def assign_operand_vector(a,b,c,d):
-        """ stub version when no vector processing included """
-        logging.info("No OVPsim vector instruction processing included")
-        if stop_on_first_error:
-            sys.exit(-1)
-    def is_an_extension_instruction(instr):
-        if 'v' in instr[0]:
-            return True
-        return False
+  def assign_operand_vector(a,b,c,d):
```
danghai: I am not sure why you have parameters `a,b,c,d`?

scripts/ovpsim_log_to_trace_csv.py
```diff
@@ -27,34 +27,35 @@
...
+    """ stub version when no vector processing included """
+    logging.info("No OVPsim vector instruction processing included")
+    if stop_on_first_error:
+      sys.exit(-1)
```
danghai: What about if we can use enum in script/lib instead of the number ?. Ex:
RET_SUCCESS = 1
RET_FAIL = 0
RET_FATAL = -1

scripts/ovpsim_log_to_trace_csv.py
udinator: Good point, changed.

scripts/ovpsim_log_to_trace_csv.py
udinator: It seems like this is a stub function until OVPsim vector instruction processing is added. @simond-imperas is this accurate?

scripts/ovpsim_log_to_trace_csv.py
udinator: Good idea, I think that might simplify things. I've added an initial implementation of this enums, PTAL

scripts/ovpsim_log_to_trace_csv.py
taoliug: I think this needs to refactored later, doesn't have to be resolved in this PR.

taoliug:
Please resolve conflict and submit this PR

simon5656:
Yes these are stub functions are currently required, with appropriate number of arguments, and are needed until the vector processing is included in this repo. [it is being develop outside due to churn]

________________________________

PR #355, opened 2019-12-12
enable nested traps

udinator:
set xSTATUS.xIE in trap handler to enable receiving nested interrupts

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -802,6 +802,21 @@ class riscv_asm_program_gen extends uvm_object;
     for (int i = 1; i < max_interrupt_vector_num; i++) begin
       string intr_handler[$];
       push_gpr_to_kernel_stack(status, scratch, cfg.mstatus_mprv, cfg.sp, cfg.tp, intr_handler);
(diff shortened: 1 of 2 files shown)


src/riscv_asm_program_gen.sv
```diff
@@ -802,6 +802,21 @@ class riscv_asm_program_gen extends uvm_object;
     for (int i = 1; i < max_interrupt_vector_num; i++) begin
       string intr_handler[$];
       push_gpr_to_kernel_stack(status, scratch, cfg.mstatus_mprv, cfg.sp, cfg.tp, intr_handler);
+      // If nested interrupts are enabled, set xSTATUS.xIE in the interrupt handler
+      // to re-enable interrupt handling capabilities
```
taoliug: Is this a spec requirement?

udinator: Yes it is, "When a trap is taken from
privilege mode y into privilege mode x, xPIE is set to the value of x IE; x IE is set to 0; and xPP
is set to y." - basically nested traps are by default disabled when taking a trap, and have to be explicitly re-enabled

PR #357, opened 2019-12-14
 Make assign_operand become a method of class RiscvInstructionTraceEntry 

danghai:
Hi
I make PR to make assign_operand become a method of class RiscvInstructionTraceEntry. Function assign_operand is going to be scale when we add new extension, and other instruction. We are going to support more methods to refactor this method.
I am not sure is it a good idea? and I do not test it yet.


taoliug:
Thanks for refactoring this

taoliug:
Sorry I have reverted this PR as it doesn't work for the coverage flow.
Please run cov.py to see if you can reproduce the problem.

danghai:
@taoliug Sorry I did not test it yet. I will try to reproduce it

PR #372, opened 2019-12-20
move handcoded asm_test generation into separate output directory

udinator:


run.py
```diff
@@ -362,20 +362,24 @@ def run_assembly(asm_test, iss_yaml, isa, mabi, iss_opts, output_dir, setting_di
...
   asm = re.sub(r"\.S$", "", asm)
   report = ("%s/iss_regr.log" % output_dir).rstrip()
-  elf = ("%s/%s.o" % (output_dir, asm))
-  binary = ("%s/%s.bin" % (output_dir, asm))
+  prefix = ("%s/directed_asm_tests/%s"  % (output_dir, asm))
+  elf = prefix + ".o"
+  binary = prefix + ".bin"
   iss_list = iss_opts.split(",")
+  run_cmd("mkdir -p %s/directed_asm_tests" % output_dir)
+  run_cmd("cp %s %s/directed_asm_tests/" % (asm_test, output_dir))
```
taoliug: One potential issue is that these hand coded tests may have dependencies to other include files, copying this file to a different directory could break this. I'd prefer to keep the hand-coded tests in the original directory, only redirect .bin/.o to directed_asm_tests directory

udinator: hmm, okay, I'll modify this accordingly

udinator: Done.

run.py
```diff
@@ -362,20 +362,23 @@ def run_assembly(asm_test, iss_yaml, isa, mabi, iss_opts, output_dir, setting_di
...
-  elf = ("%s/%s.o" % (output_dir, asm))
-  binary = ("%s/%s.bin" % (output_dir, asm))
+  prefix = ("%s/directed_asm_tests/%s"  % (output_dir, asm))
+  elf = prefix + ".o"
+  binary = prefix + ".bin"
   iss_list = iss_opts.split(",")
+  run_cmd("mkdir -p %s/directed_asm_tests" % output_dir)
   logging.info("Compiling assembly test : %s" % asm_test)
   # gcc comilation
   cmd = ("%s -static -mcmodel=medany \
```
taoliug: just a side node, we might need to support gcc_opts here to allow user pass additional options to compile their program. This could be done in a later PR.

udinator: Yep, I'm including gcc_opts in a next set of PRs to get the testlist integrated with this.

PR #388, opened 2019-12-25
Refactor riscv_trace_csv.py

danghai:
Hi @taoliug , I fixed the issue, and re-create again this PR. It looks good to me when I try to run some test cases. Could you take the look, and try to run your internal regression test cases? I remember you mention "adding a nightly regression locally". Do you have any plan for it? And is it public outside, or internal side?.
Furthermore, I am going to to use `enum` class for `TYPE` instruction. Ex:
```python
class TYPE (Enum):
  LOAD  = ['lb', 'lh', 'lw', 'lbu', 'lhu', 'ld', 'lq', 'lwu', 'ldu', 'c.lw', 'c.ld', 'c.lq', 'c.lwsp', 'c.ldsp', 'c.lqsp']
  STORE = ['sb', 'sh', 'sw', 'sd', 'sq', 'c.sw', 'c.sd', 'c.sq', 'c.swsp', 'c.sdsp', 'c.sqsp']
  CR = [ 'c.add', 'c.addw', 'c.mv', 'c.sub', 'c.and', 'c.or', 'c.xor', 'c.subw']
  ...
  ...
```
However, it still has some special case instruction. Ex:
1) I do not know how to determine the name for these groups of instructions. ['csrrw', 'csrrs', 'csrrc'], ['lui', 'auipc'], ...
2) Only one instruction. Ex: ['mv'], ['not']

Do you think is it good idea to do it? (may be an another PR if it is accepted)


taoliug:
Thanks for the update. I'd suggest keep this PR on hold until the new CSV format related changes are merged to upstream. Most of the changes in this PR are not needed after switching to new script.

danghai:
I see. Close PR

PR #391, opened 2019-12-27
integrate directed asm_tests with yaml flow

udinator:


run.py
```diff
@@ -398,25 +406,28 @@ def run_assembly(asm_test, iss_yaml, isa, mabi, iss_opts, output_dir, setting_di
...
     iss_yaml        : ISS configuration file in YAML format
     isa             : ISA variant passed to the ISS
     mabi            : MABI variant passed to GCC
+    gcc_opts        : User-defined options for GCC compilation
     iss             : Instruction set simulators
     output_dir      : Output directory of compiled test files
     setting_dir     : Generator setting directory
   """
-  result = run_cmd("find %s -name \"*.S\"" % asm_test_dir)
+  result = run_cmd("find %s -maxdepth 1 -name \"*.S\"" % asm_test_dir)
```
taoliug: why adding this constraint?

udinator: without this constraint here, the flow recursively searches every subdirectory for assembly files, which may not be intended behavior

taoliug: I think it's better to have the ability to search through multiple files.

run.py
```diff
@@ -680,29 +691,63 @@ def main():
...
   matched_list = []
+  # Any tests in the YAML test list that specify a directed assembly test
+  directed_list = []
 
   if not args.co:
     process_regression_list(args.testlist, args.test, args.iterations, matched_list, cwd)
-    if len(matched_list) == 0:
-      sys.exit("Cannot find %s in %s" % (args.test, args.testlist))
+    for t in list(matched_list):
+      if 'asm_tests' in t:
```
danghai: I think we should create a different entry name for a directed assembly test instead of using the name `test:` field in yaml. It may be name `direct_test:`. The `direct_test` does not need the field like `sim_opts, gen_opts,...`
User will not be confused between gen test, and direct assembly test. What do you think?

udinator: I don't think this would be necessary, as in the example that I added to the README in this PR, you don't need to include the `sim_opts, gen_opts, ...` to any directed assembly test entries anyways. To make this more clear, I will specify in the README that for any directed test entries that use `asm_tests` field the `gen_test` field must not be defined simultaneously.

Issue #256, opened 2019-10-24
Spike sim hanged and timeout but log file size grows 


amitgargmentor:
Hi,

if i run the spike simulation with command as shown below and if there is any bad instruction whcih cause the exception, the simulation gets timeout which is good but log file is still getting updated and grows as GB size.  

spike --isa=rv32i --log-commits -l out_2019-10-24/asm_tests/riscv_arithmetic_basic_test_0.o &> out_2019-10-24/spike_sim/riscv_arithmetic_basic_test.0.log

Have anyone faced this issue? Some time, there is stale file .nfs* created in spike_sim directory which grows in size. 

Has anyone seen this issue?

taoliug:
if you use ctrl+c to kill the spike simulation while it's running, I think the spike subprocess won't be killed based on the current implementation. However, if the spike timeout after 25s, I'd expect run.py kill the spike subprocess correctly and the log should not grow afterwards. You can use --iss_timeout=xx to set a shorter timeout period. Normally spike should finish within 5s.

amitgargmentor:
Hi ,

I didn’t use the ctrl+c and I wait for 25s for timeout ( sure, I can reduce the timeout) but the log file is still getting dumped and grow in size.

Thanks,
Amit

amitgargmentor:
I cloned the latest  repository and the spike sim doesn't grow in size after timeout . 

Thanks!!

PR #421, opened 2020-01-04
Add try-except for handling KeyboardInterrupt

danghai:
I make PR to handle KeyboardInterrupt

### DIFF
--- a/run.py
+++ b/run.py
@@ -662,54 +662,59 @@ def load_config(args, cwd):
 
 def main():
   """This is the main entry point."""
-  parser = setup_parser()
-  args = parser.parse_args()
-  cwd = os.path.dirname(os.path.realpath(__file__))
-  os.environ["RISCV_DV_ROOT"] = cwd
-  setup_logging(args.verbose)
-  # Load configuration from the command line and the configuration file.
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -662,54 +662,59 @@ def load_config(args, cwd):
...
+      if args.steps == "all" or re.match(".*iss_sim.*", args.steps):
+        iss_sim(matched_list, output_dir, args.iss, args.iss_yaml,
+                args.isa, args.core_setting_dir, args.iss_timeout)
+
+      # Compare ISS simulation result
+      if args.steps == "all" or re.match(".*iss_cmp.*", args.steps):
+        iss_cmp(matched_list, args.iss, output_dir, args.stop_on_first_error, args.exp)
+
+    sys.exit(RET_SUCCESS)
+  except KeyboardInterrupt:
```
taoliug: One question, will this also kill all the sub-processes?

danghai: I think no, you want to kill all the sub-processes?

taoliug: It's a different issue though, we can address later
https://github.com/google/riscv-dv/issues/256#issuecomment-546054136

PR #436, opened 2020-01-09
Add run_cmd_output for reporting all debug command lines

danghai:
I make PR to add `run_cmd_output` for reporting all debug command lines.

### DIFF
--- a/scripts/lib.py
+++ b/scripts/lib.py
@@ -174,6 +174,20 @@ def run_parallel_cmd(cmd_list, timeout_s = 999, exit_on_error = 0, check_return_
     os.system("stty sane")
     logging.debug(output)
 
+def run_cmd_output(cmd):
+  """Run a command and return output
+  Args:
+    cmd          : Command line to execute
+  """
+  logging.debug(" ".join(cmd))
+  try:
+    output = subprocess.check_output(cmd)
+  except subprocess.CalledProcessError as exc:
+    logging.debug(exc.output)
+    raise exc
+    sys.exit(RET_FAIL)
+  if output:
+    logging.debug(output)
 
 def process_regression_list(testlist, test, iterations, matched_list, riscv_dv_root):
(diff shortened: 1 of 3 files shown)


scripts/lib.py
```diff
@@ -174,6 +174,15 @@ def run_parallel_cmd(cmd_list, timeout_s = 999, exit_on_error = 0, check_return_
     os.system("stty sane")
     logging.debug(output)
 
+def run_cmd_output(cmd):
+  """Run a command and return output"""
+  logging.debug(" ".join(cmd))
+  try:
+    output = subprocess.check_output(cmd)
```
taoliug: I think this doesn't exactly equivalent to what was before, the output won't be printed unless there's an exception.

danghai: The reason why I do not put `logging.debug(output)` because it shows me unnecessary information. Ex:
```
Wed, 08 Jan 2020 19:12:10 lib.py:179   DEBUG    mkdir -p out_2020-01-08/asm_tests
Wed, 08 Jan 2020 19:12:10 lib.py:186   DEBUG    b''
```

taoliug: I think we could do some filtering about the redundant b'', but we should keep the capability of show the verbose log. The verbose log at the instruction generation phase is useful for debugging.

danghai: What do you mean instruction generation phase? I add `debug_log` allow to showing the verbose log by default.

taoliug: I mean when we compile the generator and run the RTL simulator to generate the test. After we get to iss_sim stage, the log is redirected to a file, so meaningful screen log will be printed.  The change looks good to me. BTW, have you try something like this?

```
if output:
  logging.debug(output)
```

Just wonder if we can omit the print when the output is actually empty.

danghai: Yes. It is better

PR #453, opened 2020-01-23
Support running C test

danghai:
I create PR to support C test. It looks the same as running assembly test except the option, file on command. I will enhance after it works. I try to write a simple C file, Unfortunately, it does not work.
Example:
$ run --c_test test.c --debug debug.log -o c_test

Debug command:
```
/opt/riscv/bin/riscv64-unknown-elf-gcc -mcmodel=medany -nostdlib -nostartfiles test.c -I/home/danghai/hhoangda/riscv-dv/user_extension -T/home/danghai/hhoangda/riscv-dv/scripts/link.ld -o c_test/directed_c_tests/test.o -march=rv32imc -mabi=ilp32

/opt/riscv/bin/riscv64-unknown-elf-objcopy -O binary c_test/directed_c_tests/test.o c_test/directed_c_tests/test.bin

/opt/riscv/bin/spike --log-commits --isa=rv32imc -l c_test/directed_c_tests/test.o &> c_test/spike_sim/test.log

```

### DIFF
--- a/run.py
+++ b/run.py
@@ -440,6 +440,93 @@ def run_assembly_from_dir(asm_test_dir, iss_yaml, isa, mabi, gcc_opts, iss,
   else:
     logging.error("No assembly test(*.S) found under %s" % asm_test_dir)
 
+def run_c(c_test, iss_yaml, isa, mabi, gcc_opts, iss_opts, output_dir,
+          setting_dir, debug_cmd):
+  """Run a directed c test with ISS
+
+  Args:
+    c_test      : C test file
+    iss_yaml    : ISS configuration file in YAML format
+    isa         : ISA variant passed to the ISS
+    mabi        : MABI variant passed to GCC
+    gcc_opts    : User-defined options for GCC compilation
+    iss_opts    : Instruction set simulators
+    output_dir  : Output directory of compiled test files
+    setting_dir : Generator setting directory
+    debug_cmd   : Produce the debug cmd log without running
+  """
+  if not c_test.endswith(".c"):
+    logging.error("%s is not a .c file" % c_test)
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -440,6 +440,93 @@ def run_assembly_from_dir(asm_test_dir, iss_yaml, isa, mabi, gcc_opts, iss,
...
+    mabi            : MABI variant passed to GCC
+    gcc_opts        : User-defined options for GCC compilation
+    iss             : Instruction set simulators
+    output_dir      : Output directory of compiled test files
+    setting_dir     : Generator setting directory
+    debug_cmd       : Produce the debug cmd log without running
+  """
+  result = run_cmd("find %s -name \"*.c\"" % asm_test_dir)
+  if result:
+    asm_list = result.splitlines()
```
taoliug: should this be change to c_list? here and below

run.py
```diff
@@ -440,6 +440,93 @@ def run_assembly_from_dir(asm_test_dir, iss_yaml, isa, mabi, gcc_opts, iss,
...
+    logging.info("Found %0d assembly tests under %s" %
+                 (len(asm_list), asm_test_dir))
+    for asm_file in asm_list:
+      run_assembly(asm_file, iss_yaml, isa, mabi, gcc_opts, iss, output_dir,
```
taoliug: should this be run_c?

danghai: Oh I do not change it yet. Because `run_c` does not work. Could you check the options on gcc command?

taoliug: That's fine, we can have a follow up PR to make it work.

taoliug:
update #453

PR #472, opened 2020-02-05
Add pmp configuration object

udinator:
- added some new randomization macros for ease of use
- the uvm_cmdline_processor instance has been moved to the riscv_instr_pkg to allow the command line parsing functions to be used by various configuration objects as needed
- basic pmp configuration object added - with just the default values for now

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class riscv_pmp_cfg extends uvm_object;
+
+  // default to a single PMP region
+  rand int pmp_num_regions = 1;
+  // default to granularity of 0 (4 bytes grain)
+  rand int pmp_granularity = 0;
+  // enable bit for pmp randomization
+  bit pmp_randomize = 0;
+  // pmp CSR configurations
+  rand pmp_cfg_reg_t pmp_cfg[];
+
+  // used to parse addr_mode configuration from cmdline
+  typedef uvm_enum_wrapper#(pmp_addr_mode_t) addr_mode_wrapper;
+  pmp_addr_mode_t addr_mode;
+
+  `uvm_object_utils_begin(riscv_pmp_cfg)
+    `uvm_field_int(pmp_num_regions, UVM_DEFAULT)
+    `uvm_field_int(pmp_granularity, UVM_DEFAULT)
+  `uvm_object_utils_end
+
+  // constraints
+  constraint sanity_c {
+    pmp_num_regions inside {[1 : 16]};
+    pmp_granularity inside {[0 : XLEN + 3]};
+  }
+
+  // TODO(udinator) move these constraints to post_randomize() to save performance
+  constraint xwr_c {
+    foreach (pmp_cfg[i]) {
+      !(pmp_cfg[i].w && !pmp_cfg[i].r);
+    }
+  }
+
+  constraint grain_addr_mode_c {
+    foreach (pmp_cfg[i]) {
+      (pmp_granularity >= 1) -> (pmp_cfg[i].a != NA4);
+    }
+  }
+
+  function new(string name = "");
+    string s;
+    super.new(name);
+    get_bool_arg_value("+pmp_randomize=", pmp_randomize);
+    pmp_cfg = new[pmp_num_regions];
+    if (!pmp_randomize) begin
+      set_defaults();
+    end
+  endfunction
+
+  // TODO(udinator) partition address space to map to all active pmp_addr CSRs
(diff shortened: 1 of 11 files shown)


src/riscv_pmp_cfg.sv
```diff
@@ -0,0 +1,142 @@
...
+        end
+      end else if (pmp_cfg[i].a inside {OFF, TOR}) begin
+        for (int j = 0; j <= pmp_granularity - 1; j = j + 1) begin
+          pmp_cfg[i].addr[j] = 1'b1;
+        end
+      end
+    end
+  endfunction
+
+  function void gen_pmp_instr(ref string instr[$], riscv_reg_t scratch_reg);
```
taoliug: Can you add some high level description about this function?

src/riscv_pmp_cfg.sv
```diff
@@ -0,0 +1,142 @@
...
+    `uvm_field_int(pmp_granularity, UVM_DEFAULT)
+  `uvm_object_utils_end
+
+  // constraints
+  constraint sanity_c {
+    pmp_num_regions inside {[1 : 16]};
+    pmp_granularity inside {[0 : XLEN + 3]};
+  }
+
+  // TODO(udinator) move these constraints to post_randomize() to save performance
```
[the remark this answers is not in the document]
udinator: Currently no, since I'm only enabling the default values, it's just something to keep in my mind for when I start enabling the max number of regions.

src/riscv_pmp_cfg.sv
```diff
@@ -0,0 +1,142 @@
...
+      (pmp_granularity >= 2 && pmp_cfg[i].a[1]) -> (pmp_cfg[i].a == NAPOT);
+      (pmp_granularity >= 1 && !pmp_cfg[i].a[1]) -> (pmp_cfg[i].a inside {[OFF : TOR]});
+    }
+  }
+
+  function new(string name = "");
+    string s;
+    super.new(name);
+    get_bool_arg_value("+pmp_randomize=", pmp_randomize);
+    pmp_cfg = new[pmp_num_regions];
```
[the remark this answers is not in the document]
udinator: Yes, this PR is just to use the default pmp values, it will be slightly modified once I fill in the skeleton code with full configurability.

src/riscv_pmp_cfg.sv
```diff
@@ -0,0 +1,142 @@
...
+    string pmp_region;
+    get_int_arg_value("+pmp_num_regions=", pmp_num_regions);
+    get_int_arg_value("+pmp_granularity=", pmp_granularity);
+    // TODO(udinator) - parse the pmp configuration values
+  endfunction
+
+  function void post_randomize();
+    foreach (pmp_cfg[i]) begin
+      if (pmp_cfg[i].a == NAPOT) begin
+        for (int j = 0; j <= pmp_granularity - 2; j = j + 1) begin
```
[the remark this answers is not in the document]
udinator: I actually have deleted this function in a later commit to this PR.
Essentially, PMP mandates some read masks for the pmpaddr CSRs when in certain address matching modes, but it seems that hardware should handle this, and that software should not be responsible for doing this masking, like I was doing in this function.

src/riscv_pmp_cfg.sv
udinator: Done

src/riscv_pmp_cfg.sv
```diff
@@ -0,0 +1,130 @@
...
+    foreach (pmp_cfg[i]) begin
+      pmp_id = i / cfg_per_csr;
+      cfg_byte = {pmp_cfg[i].l, pmp_cfg[i].zero, pmp_cfg[i].a,
+                  pmp_cfg[i].x, pmp_cfg[i].w, pmp_cfg[i].r};
+      `uvm_info(`gfn, $sformatf("cfg_byte: 0x%0x", cfg_byte), UVM_DEBUG)
+      cfg_bitmask = cfg_byte << ((i % cfg_per_csr) * 8);
+      `uvm_info(`gfn, $sformatf("cfg_bitmask: 0x%0x", cfg_bitmask), UVM_DEBUG)
+      pmp_word = pmp_word | cfg_bitmask;
+      `uvm_info(`gfn, $sformatf("pmp_word: 0x%0x", pmp_word), UVM_DEBUG)
+      cfg_bitmask = 0;
```
taoliug: this can be rewrite to pmp_word = ( pmp_word << 8 ) | cfg_byte ?

src/riscv_pmp_cfg.sv
```diff
@@ -0,0 +1,130 @@
...
+      //TODO (udinator) - add rv64 support for pmpaddr writes
+      instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg, pmp_cfg[i].addr[XLEN + 1 : 2]));
+      instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmp_addr + i, scratch_reg));
+      // short circuit if end of list
+      if (i == pmp_cfg.size() - 1) begin
+        instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg, pmp_word));
+        instr.push_back($sformatf("csrw 0x%0x, x%0d",
+                                  base_pmpcfg_addr + (i / cfg_per_csr),
```
taoliug: i / cfg_perf_csr -> pmp_id, here can below

udinator: good catch, thanks for pointing that out

src/riscv_pmp_cfg.sv
udinator: I can't do that for this situation, since the physical pmpcfg CSR is arranged like this:
pmpcfg0 = {cfg3, cfg2, cfg1, cfg0} in reverse order from MSB to LSB, so I have to left-shift all the new config bytes to OR with the pmp_word.
I'll leave it as this current calculation for now, but will take a TODO item to try and find a more condensed calculation.

PR #476, opened 2020-02-12
The "solve...before" constraints should be modified.

dariuszst:
According to the SV 1800-2017 standard using non random variable as a solve...before argument is not allowed. In two cases - for variable "init_privileged_mode" and "use_sp_as_rs1 " the randomization possibilities are disabled by rand_mode(0) but the "solve...before" constraint is still active causing compilation problems. To fix that the "solve...before" constraint should be divided and the ordering part should be disabled before randomization. After randomization, if it is necessary the "solve...before" constraint can be re-enabled.

Kind regards,
Darek

### DIFF
--- a/src/riscv_instr_gen_config.sv
+++ b/src/riscv_instr_gen_config.sv
@@ -379,8 +379,11 @@ class riscv_instr_gen_config extends uvm_object;
     unique {gpr};
   }
 
-  constraint addr_translaction_c {
+  constraint addr_translaction_rnd_order_c {
     solve init_privileged_mode before virtual_addr_translation_on;
+  }
+  
+  constraint addr_translaction_c {
     if ((init_privileged_mode != MACHINE_MODE) && (SATP_MODE != BARE)) {
       virtual_addr_translation_on == 1'b1;
     } else {
@@ -517,6 +520,7 @@ class riscv_instr_gen_config extends uvm_object;
                   $sformatf("Illegal boot mode option - %0s", boot_mode_opts))
       endcase
       init_privileged_mode.rand_mode(0);
+      addr_translaction_rnd_order_c.constraint_mode(0);
     end
     `uvm_info(`gfn, $sformatf("riscv_instr_pkg::supported_privileged_mode = %0d",
                    riscv_instr_pkg::supported_privileged_mode.size()), UVM_LOW)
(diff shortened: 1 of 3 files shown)


taoliug:
Thanks for your contribution, can you sign the CLA and reply "@googlebot I signed it" here?

dariuszst:
Hello,
I signed the CLA. Sorry for the delay.

taoliug:
Can you reply like this?
@googlebot I signed it

src/riscv_instr_gen_config.sv
```diff
@@ -379,8 +379,11 @@ class riscv_instr_gen_config extends uvm_object;
     unique {gpr};
   }
 
-  constraint addr_translaction_c {
+  constraint addr_translaction_rnd_order_c {
     solve init_privileged_mode before virtual_addr_translation_on;
```
taoliug: they are both random variables, why this change is needed?

src/riscv_load_store_instr_lib.sv
```diff
@@ -39,8 +39,11 @@ class riscv_load_store_base_instr_stream extends riscv_mem_access_stream;
 
   `uvm_object_utils(riscv_load_store_base_instr_stream)
 
-  constraint sp_c {
+  constraint sp_rnd_order_c {
     solve use_sp_as_rs1 before rs1_reg;
```
taoliug: same here

src/riscv_instr_gen_config.sv
dariuszst: The "init_privileged_mode " is not random since you disable "random" few lines below (line 519)

src/riscv_instr_gen_config.sv
```diff
@@ -517,6 +520,7 @@ class riscv_instr_gen_config extends uvm_object;
                   $sformatf("Illegal boot mode option - %0s", boot_mode_opts))
       endcase
       init_privileged_mode.rand_mode(0);
```
dariuszst: If the boot mode option is specified then the random attribute of init_privileged_mode is disabled by  init_privileged_mode.rand_mode(0); line - thus the constraint "addr_translaction_c" cannot be solved.

src/riscv_load_store_instr_lib.sv
dariuszst: The "use_sp_as_rs1" is not "rand" because it is disabled in the pre_randomize() function in line 99: use_sp_as_rs1.rand_mode(0);

taoliug:
Thanks for the fix

Issue #443, opened 2020-01-17
Add coverage model for floating point extension


taoliug:
File this bug to track the implementation progress of floating point coverage model. A few things to start with:
- Cover all possible operands for each floating point instruction
- Numeric corner cases for the arithmetic instructions (NaN, subnormal values etc)
- Floating point CSR coverage (fcsr)

taoliug:
Weicai, can you give a summary what's left to be done for this task? Thanks

weicaiyang:
> Weicai, can you give a summary what's left to be done for this task? Thanks

@taoliug 
These are done
1. single precision 11.1-11.6 sections
2. parsing instruction should work for all floating point instructions

Remaining:
1. 11.7-11.9
2. chapter 12 double precision
3. floating point special values

If you're busy on integrating vector extension, I continue to this one.

weicaiyang:
with #615 merged, use more meaningful initial value for floating point registers and I see most of the floating point coverage are 100%. 
For instructions like FMADD, rd=rs1×rs2+rs3, it's hard to cover all kinds of special values for rd. Perhaps we need some direct tests or just remove the coverage point for rd, since we have cover all the other operands.

taoliug:
Closing this bug as all related work is completed

PR #488, opened 2020-02-25
Add coverage for single precision floating (Part 1)

weicaiyang:
Implemented section 11.1-11.7 and will create another PR the remaining
instructions


src/riscv_instr_cover_group.sv
```diff
@@ -446,6 +471,57 @@ class riscv_instr_cover_group;
...
+  `SFP_R_INSTR_CG_BEGIN(fmin_s)
+    cp_sign_cross: cross cp_rs1_sign, cp_rs2_sign;
+  `CG_END
+
+  `SFP_R_INSTR_CG_BEGIN(fmax_s)
+    cp_sign_cross: cross cp_rs1_sign, cp_rs2_sign;
+  `CG_END
+
+  `SFP_R_INSTR_CG_BEGIN(fmadd_s)
+    cp_sign_cross: cross cp_rs1_sign, cp_rs2_sign, cp_rd_sign;
```
taoliug: do you need to cover rs3 for fmadd, fnmadd, fmsub, fnmsub?

src/riscv_instr_cover_group.sv
```diff
@@ -446,6 +471,57 @@ class riscv_instr_cover_group;
...
+  `CG_END
+
+  `SFP_R_INSTR_CG_BEGIN(fnmadd_s)
+    cp_sign_cross: cross cp_rs1_sign, cp_rs2_sign, cp_rd_sign;
+  `CG_END
+
+  `SFP_R_INSTR_CG_BEGIN(fmsub_s)
+    cp_sign_cross: cross cp_rs1_sign, cp_rs2_sign, cp_rd_sign;
```
taoliug: can you confirm if all these combinations are reachable?

taoliug:
update #443

src/riscv_instr_cover_group.sv
weicaiyang: Good catch. And it's R4 format rather than R format. I also use fs1/2/3 instead of rs1/2. PTAL, thanks

src/riscv_instr_cover_group.sv
weicaiyang: I feel last operation is sub/add, so it can cross all the signs of fs1,fs2,fs3,fd

weicaiyang:
@taoliug Please take a look

taoliug:
update #443

PR #504, opened 2020-03-05
Don't change input file in spike_log_to_trace_csv.py

rswarbrick:
Yesterday, I spent half an hour getting very confused about why my
Spike logs (which didn't some magic lines) were getting replaced with
the empty file when I tried to use them as inputs to this script.

This patch replaces the 'sed -i' invocations with equivalent logic in
the actual Python. It also factors out the code that parses the trace
from the code that uses it, which I think makes things a bit more
obvious.

Finally, the patch adds lots of comments, explaining the control flow:
when writing this, I really struggled to figure out what the existing
code was trying to do with things like WFI and ECALL instructions and
architectural updates. Hopefully, this is much clearer now.

Other than not rewriting input files, there are two known changes
caused by this patch. Firstly, the previous code wrote out the jr
instruction at 0x1010, but only if called with full_trace=1. I don't
believe that can have been intentional, so I've picked the behaviour
that is used when we compare traces in Ibex.

Secondly, the instruction counting has changed. I think that the
counting previously included the jump at 0x1010 (which we now don't
write out). We don't include that any more, so the "Processed
instruction count" message will decrease by 1.

### DIFF
--- a/scripts/spike_log_to_trace_csv.py
+++ b/scripts/spike_log_to_trace_csv.py
@@ -34,70 +34,6 @@ ILLE_RE  = re.compile(r"trap_illegal_instruction")
 
 LOGGER = logging.getLogger()
 
-def process_spike_sim_log(spike_log, csv, full_trace = 0):
-  """Process SPIKE simulation log.
-
-  Extract instruction and affected register information from spike simulation
-  log and save to a list.
-  """
-  logging.info("Processing spike log : %s" % spike_log)
-  instr_cnt = 0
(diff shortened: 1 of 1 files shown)


rswarbrick:
@googlebot I signed it!

rswarbrick:
For any reviewer: the force push is just to fix a silly typo in the commit message.

taoliug:
Thanks for the refactoring. Tried it locally and it works well.

PR #505, opened 2020-03-05
improve documentation of config options

udinator:


docs/source/configuration.rst
```diff
@@ -145,57 +145,75 @@ You can also add directed assembly/C test in the testlist
...
++---------------------------------+---------------------------------------------------+---------+
+| no_wfi                          | Disable WFI instruction                           | 1       |
++---------------------------------+---------------------------------------------------+---------+
+| set_mstatus_tw                  | Enable WFI to be treated as illegal instruction   | 0       |
++---------------------------------+---------------------------------------------------+---------+
+| no_dret                         | Disable dret instruction                          | 1       |
++---------------------------------+---------------------------------------------------+---------+
+| no_branch_jump                  | Disable branch/jump instruction                   | 0       |
++---------------------------------+---------------------------------------------------+---------+
+| no_load_store                   | Disable load/store instruction                    | 0       |
```
taoliug: should this be removed since it's not used?

docs/source/configuration.rst
```diff
@@ -145,57 +145,75 @@ You can also add directed assembly/C test in the testlist
 
 Runtime options of the generator
```
taoliug: can you run the document generator to update the generated html files

docs/source/configuration.rst
udinator: Good point, I forgot to get rid of it.

PR #517, opened 2020-03-13
Add bit manipulation

weicaiyang:
Integrate this from PR #498 added by @adimay162
Also run with formatter(verible/verilog_format) and add license header


weicaiyang:
it doesn't work. Let me try another way.

PR #519, opened 2020-03-14
[cov] tag coverage database directories with <test_id>

udinator:
Renames coverage database output directories to `test_<test_id>` from `test_<seed>` to provide uniquification of names for use cases where a same seed value is passed to every test iteration (like in Ibex environment).

### DIFF
--- a/yaml/simulator.yaml
+++ b/yaml/simulator.yaml
@@ -29,7 +29,7 @@
     cmd: >
       <out>/vcs_simv +vcs+lic+wait <sim_opts> +ntb_random_seed=<seed> <cov_opts>
     cov_opts: >
-      -cm_dir <out>/test.vdb -cm_log /dev/null -cm_name test_<seed>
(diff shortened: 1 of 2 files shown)


yaml/simulator.yaml
```diff
@@ -29,7 +29,7 @@
     cmd: >
       <out>/vcs_simv +vcs+lic+wait <sim_opts> +ntb_random_seed=<seed> <cov_opts>
     cov_opts: >
-      -cm_dir <out>/test.vdb -cm_log /dev/null -cm_name test_<seed>
+      -cm_dir <out>/test.vdb -cm_log /dev/null -cm_name test_<test_id>
```
taoliug: Do you need to keep the seed? If the test is run multiple times with different seeds, this will end up overriding the same coverage directory.

udinator: Good point, I've appended the seed to this tag as well

PR #523, opened 2020-03-20
Update style check

weicaiyang:
1. Add to run style check when you run sim
2. Update codes for new rule below in latest style tool
 - Explicitly define a default case for every case statement
3. Update exclude_filelist as a tool bug has been fixed

Everyone needs to run ./verilog_style/build-verible.sh to get the latest
version

Signed-off-by: Weicai Yang <weicai@google.com>

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -106,7 +106,8 @@ class riscv_asm_program_gen extends uvm_object;
       `DV_CHECK_RANDOMIZE_FATAL(main_program[hart])
       main_program[hart].gen_instr(.is_main_program(1), .no_branch(cfg.no_branch_jump));
       // Setup jump instruction among main program and sub programs
-      gen_callstack(main_program[hart], sub_program[hart], sub_program_name, cfg.num_of_sub_program);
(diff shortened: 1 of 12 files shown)


src/riscv_asm_program_gen.sv
```diff
@@ -1161,6 +1168,7 @@ class riscv_asm_program_gen extends uvm_object;
           USTATUS: begin
             interrupt_handler_instr.push_back($sformatf("csrsi 0x%0x, 0x%0x", status, 1));
           end
+          default: `uvm_info(`gfn, $sformatf("Unsupported status %0s", status), UVM_LOW)
```
taoliug: uvm_fatal

src/riscv_asm_program_gen.sv
```diff
@@ -821,6 +826,8 @@ class riscv_asm_program_gen extends uvm_object;
             gen_trap_handler_section(hart, "u", UCAUSE, UTVEC, UTVAL,
                                      UEPC, USCRATCH, USTATUS, UIE, UIP);
           end
+        default: `uvm_info(`gfn, $sformatf("Unsupported privileged_mode %0s",
```
taoliug: uvm_fatal

Issue #520, opened 2020-03-14
Support Bit-manipulation extension


taoliug:
File this bug to track the progress of Bit-manipulation extension support. 
Thanks for the contribution from Mellanox, we have got the basic Bit-manipulation instruction class ready (#518). Following work:
- Create a target for B-extension
- Add tests for B-extensions. 
- [Get ovpsim setup and pass ISS sim.](https://github.com/riscv/riscv-bitmanip#imperas-riscvovpsim-reference-simulation)
- Update document
- Add functional coverage model for B-extension

Please leave your comment here if you have suggestion or feature request on B-extension testing.

taoliug:
A rv32imcb target is created, OVPsim setting is also updated.

taoliug:
I believe below steps are done already.

- Create a target for B-extension
- Add tests for B-extensions.
- Get ovpsim setup and pass ISS sim.

weicaiyang:
> I believe below steps are done already.
> 
> * Create a target for B-extension
> * Add tests for B-extensions.
> * Get ovpsim setup and pass ISS sim.

Yes, these are done.

taoliug:
Add one item: Enable spike & ovpsim co-simulation flow for B-extension.
When you update the document, please provide the links for the required toolchain

taoliug:
Closing this bug as all required work is done. Thanks Weicai!

PR #526, opened 2020-03-23
Update b-extention

weicaiyang:
1. Add property constraint for imm value
2. Fix misplacement of some 64B instructions
3. Fix function riscv_b_instr.convert2asm

Signed-off-by: Weicai Yang <weicai@google.com>


target/rv32imcb/testlist.yaml
```diff
@@ -40,6 +40,7 @@
   gen_test: riscv_rand_instr_test
   gen_opts: >
     +disable_compressed_instr=1
+    +enable_b_extension=1
```
taoliug: Please create a new test entry for bit manipulation rather than changing existing test

weicaiyang: Updated, PTAL. Thanks

taoliug:
update #520

PR #527, opened 2020-03-25
configurable mtvec alignment

udinator:


target/ml/riscv_core_setting.sv
```diff
@@ -35,6 +35,10 @@ riscv_instr_group_t supported_isa[$] = {RV32I, RV32M, RV64I, RV64M, RV32C, RV64C
 // Interrupt mode support
 mtvec_mode_t supported_interrupt_mode[$] = {DIRECT, VECTORED};
 
+// *TVEC CSR alignment
+// This value should be set to log_2 of the intended byte alignment of the *tvec base field
+int tvec_alignment = 2;
```
taoliug: can you move it to riscv_instr_gen_config so that it doesn't impact existing user code?

udinator: Sure, I initially felt this was something that is more on a per-core basis so I put it in the core_setting file, but I can move it to the cfg.

udinator: Done

PR #530, opened 2020-03-31
Change dynamic array parameters to const.

danielmlynek:
Parameters should be constant at elaboration time. Creating parameter of dynamic array type (which main idea it to be dynamic at runtime) is not clearly defined as legal by LRM. 

Please change it to a const array - it should express expected behaviour better.


danielmlynek:
@googlebot I signed it!

dariuszst:
@googlebot I signed it!

dariuszst:
@googlebot I consent.

PR #532, opened 2020-04-06
Add options to enable bitmanip by group

weicaiyang:
Also add a test that enables ZBB only

### DIFF
--- a/src/isa/riscv_b_instr.sv
+++ b/src/isa/riscv_b_instr.sv
@@ -470,6 +470,51 @@ class riscv_b_instr extends riscv_instr;
     this.has_rs3 = rhs_.has_rs3;
   endfunction : do_copy
 
+  virtual function bit is_supported(riscv_instr_gen_config cfg);
+    return cfg.enable_b_extension && (
+           (ZBB inside {cfg.enable_bitmanip_groups} && instr_name inside {
+               CLZ, CTZ, PCNT,
+               SLO, SLOI, SLOW, SLOIW,
+               SRO, SLOI, SROW, SLOIW,
+               MIN, MINU, MAX, MAXU,
+               ADDWU, ADDIWU, SUBWU,
+               ADDU_W, SUBU_W,
+               SLLIU_W,
+               ANDN, ORN,
+               XNOR, PACK, PACKW, PACKU, PACKUW, PACKH,
+               ROL, ROLW, ROR, RORW, RORI, RORIW
+               }) ||
+           (ZBS inside {cfg.enable_bitmanip_groups} && instr_name inside {
+               SBSET, SBSETW, SBSETI, SBSETIW,
+               SBCLR, SBCLRW, SBCLRI, SBCLRIW,
+               SBINV, SBINVW, SBINVI, SBINVIW,
+               SBEXT, SBEXTW, SBEXTI
+               }) ||
+           (ZBP inside {cfg.enable_bitmanip_groups} && instr_name inside {
+               GREV, GREVW, GREVI, GREVIW,
+               GORC, GORCW, GORCI, GORCIW,
+               SHFL, SHFLW, UNSHFL, UNSHFLW, SHFLI, UNSHFLI
+               }) ||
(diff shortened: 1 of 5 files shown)


src/isa/riscv_instr.sv
```diff
@@ -135,6 +135,42 @@ class riscv_instr extends uvm_object;
     create_csr_filter(cfg);
   endfunction : create_instr_list
 
+  static function bit is_supported_b_ext_instr(riscv_instr_gen_config cfg,
```
taoliug, line 138: Can you find a way to refactor this function to riscv_b_instr class?  We could have a function like is_supported_instr in the riscv_instr class. In the riscv_b_instr class, just implement it based on the enabled instructions.

src/riscv_instr_gen_config.sv
```diff
@@ -239,6 +239,15 @@ class riscv_instr_gen_config extends uvm_object;
...
   bit                    enable_b_extension;
+  bit                    enable_bitmanip_zbb;
+  bit                    enable_bitmanip_zbs;
+  bit                    enable_bitmanip_zbp;
+  bit                    enable_bitmanip_zbe;
+  bit                    enable_bitmanip_zbf;
+  bit                    enable_bitmanip_zbc;
+  bit                    enable_bitmanip_zbr;
+  bit                    enable_bitmanip_zbm;
+  bit                    enable_bitmanip_zbt;
```
taoliug, line 250: I'd suggest to simplify this as enabled_bitmanip_groups[$] = {"zbb", "..."}; I think this is a static configuration, you don't need command line options to override it.

udinator: Could we add this `enabled_bitmanip_groups` list to the `riscv_core_setting.sv` file?

taoliug: I think we can keep it in the cfg class, I don't want to impact user code since this setting is not relevant for most users.

src/isa/riscv_instr.sv
weicaiyang: can't use virtual and static together. We need to make `create_instr_list`non-static in order to refactor it. WDYT?
>   Invalid use of class item qualifiers. Cannot use virtual and static keywords
>   together for method declarations.

taoliug: Can you try to add something after line 113?

instr_inst = create_instr(instr_name);
if  (!instr_inst.is_supported(cfg)) continue;
We could make is_supported a virtual function.

weicaiyang: make sense. It works. Thanks

src/riscv_instr_gen_config.sv
weicaiyang: Updated, PTAL

src/isa/riscv_b_instr.sv
```diff
@@ -470,6 +470,51 @@ class riscv_b_instr extends riscv_instr;
     this.has_rs3 = rhs_.has_rs3;
   endfunction : do_copy
 
+  virtual function bit is_supported(riscv_instr_gen_config cfg);
+    return cfg.enable_b_extension ||
```
taoliug, line 474: should this be &&?

weicaiyang: `enable_b_extension` enables all B extension
`enable_bitmanip_groups` enables one/some of the groups
I feel this way may be easy to use. or I can rename it to `enable_all_b_extension`

taoliug: With the current naming, it's werid to have cfg.enable_b_extension = 0 but still generating b-ext instructions. If you go with enable_all_b_extension approach, I'd suggest to just remove enable_all_b_extension and have the default value of enable_bitmanip_groups to include all groups. It's better to simplify the options.

weicaiyang: OK, I will keep `enable_bitmanip_groups` to enable/disable b-ext and set  `enable_bitmanip_groups` to include all groups by default

PR #536, opened 2020-04-08
fix pmp randomization to adhere to max offset

udinator:
This constrains randomization of pmp address to all still lie within the address range defined by:
```systemverilog
[<main> : <main> + pmp_max_offset]
```

Additionally an enable flag is added to control whether randomized addresses can overlap.

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -25,6 +25,9 @@ class riscv_pmp_cfg extends uvm_object;
   // enable bit for pmp randomization
   bit pmp_randomize = 0;
 
+  // allow pmp randomization to cause address range overlap
+  bit pmp_allow_addr_overlap = 0;
+
   // pmp CSR configurations
   rand pmp_cfg_reg_t pmp_cfg[];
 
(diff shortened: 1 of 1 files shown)


diop1:
Hi, 
I just clone  the risc-dv and try to run one test : python3 run.py -si ius

my cmd :  python3 run.py -si ius
parsing error  on  : cmdline_enum_processor 

Wed, 15 Apr 2020 21:37:32 INFO     Building RISC-V instruction generator 
               class cmdline_enum_processor #(parameter type T);
                                                 |
xmvlog: *E,SVTMAP (/home/bara/ips_verif/tests/riscv/riscv-dv/src/riscv_instr_pkg.sv,1249|49): Omitting the data_type from a type_assignment in a parameter_port_list currently is not supported.
    static function void get_array_values(string cmdline_str, ref T vals[]);
                                                                  |
xmvlog: *E,SVNOTY (/home/bara/ips_verif/tests/riscv/riscv-dv/src/riscv_instr_pkg.sv,1250|66): Syntactically this identifier appears to begin a datatype but it does not refer to a visible datatype in the 

Did I miss something ?

udinator:
Hi @diop1 - Thanks for raising this issue. If it's all right with you, I am going to copy over your comment into a new GitHub issue to allow for easier discussion there.

PR #537, opened 2020-04-09
Add config knob to add 4 to mepc

udinator:
For situations like PMP, we need to add 4 to MEPC during all relevant exception handlers, otherwise it will cause an infinite loop of trying to execute the same instruction, repeating the same PMP fault over and over again.

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -601,7 +601,7 @@ class riscv_asm_program_gen extends uvm_object;
 
   virtual function void pre_enter_privileged_mode(int hart);
     string instr[];
-	string str[$];    
+	string str[$];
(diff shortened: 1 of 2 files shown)


taoliug:
How do we know MEPC + 4 is a valid instruction boundary?

udinator:
For the time being, this option can only be used if compressed instructions are disabled, I don't have a way right now to be able to tell if a particular instruction is compressed or not.

taoliug:
I think one possible way to solve this problem is trying to fix the PMP configure one at a time. Eventually all PMP regions will be fixed and test should go through.

udinator:
I think you're right, I'd initially planned on adding that feature a bit later, but I think it might be necessary to have now, I'll close this PR for the time being and re-open a new one with this new set of changes.

Issue #538, opened 2020-04-15
IUS compile errors


udinator:
From user @diop1 in #536:

```console
Hi,
I just clone the risc-dv and try to run one test : python3 run.py -si ius

my cmd : python3 run.py -si ius
parsing error on : cmdline_enum_processor

Wed, 15 Apr 2020 21:37:32 INFO Building RISC-V instruction generator
class cmdline_enum_processor #(parameter type T);
|
xmvlog: *E,SVTMAP (/home/bara/ips_verif/tests/riscv/riscv-dv/src/riscv_instr_pkg.sv,1249|49): Omitting the data_type from a type_assignment in a parameter_port_list currently is not supported.
static function void get_array_values(string cmdline_str, ref T vals[]);
|
xmvlog: *E,SVNOTY (/home/bara/ips_verif/tests/riscv/riscv-dv/src/riscv_instr_pkg.sv,1250|66): Syntactically this identifier appears to begin a datatype but it does not refer to a visible datatype in the

Did I miss something ?
```

diop1:
fixed. thanks

PR #535, opened 2020-04-07
Initial Aldec Riviera-Pro simulator support

danielmlynek:
Changes in original code are are made to make it compatible with Riviera-PRO.
Some changes have comments to make it clear why they are needed.

I closed previous pull request on this topic to clean the history.

PR #539, opened 2020-04-15
Fix ius compile error

weicaiyang:
Fix for #538
1. add default parameter type for cmdline_enum_processor
2. update change from #535

### DIFF
--- a/src/riscv_instr_pkg.sv
+++ b/src/riscv_instr_pkg.sv
@@ -1046,17 +1046,9 @@ package riscv_instr_pkg;
   // TODO (udinator) allow a full 34 bit address for rv32?
 `ifdef _VCP //GRK958
   typedef struct packed {
-    bit                   l;
-    bit [1:0]                  zero;
-    pmp_addr_mode_t       a;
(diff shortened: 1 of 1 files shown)


src/riscv_instr_pkg.sv
```diff
@@ -1046,17 +1046,9 @@ package riscv_instr_pkg;
   // TODO (udinator) allow a full 34 bit address for rv32?
 `ifdef _VCP //GRK958
```
weicaiyang, line 1047: @danielmlynek I made some change on your recent update to shorten the code and added `endif`. Please let me know if this doesn't work for Aldec Riviera-Pro

danielmlynek: I'm sorry - I've forgotten add endif when doing fast upmerge to master.
I thought that anyway this code before merging in will be checked by CI  (ie after my wrong commit)- so Im suprised that this issue was detected after merge.
As per yours changes - rand members cannot be done in packed struct - it will fail on compilation.
I will open separate PR to fix this.

PR #546, opened 2020-04-20
update PMP exception handlers to 'fix' config CSRs

udinator:
This PR adds new functionality to the PMP configuration object to enable generation of a custom PMP fault handling routine, that will loop through each `pmpaddr/pmpcfg` CSR pair, match up the correct CSRs to the faulting address, then "fix" the configuration settings by enabling the appropriate access permissions.

For example, say `pmpaddr0` is set to `0x80002000`, and its corresponding 8-bit access configuration fields are set such that the address matching scheme is `TOR` and all instruction fetches are disabled. If an instruction fetch to `0x80001FFC` is made, a PMP fault will occur, and the core will jump to the generated instruction fetch exception handler. Assuming `mepc` is not modified, after executing the handler the core will jump back to the same address that caused the trap, which means that we will infinitely loop on this instruction fetch until any timeouts.

This approach was chosen to allow forward progress as simply updating `mepc` within the handler itself is not desirable as there is no good way of knowing if the faulting instruction is compressed or not (which directly translates into whether we add 2 or 4 to `mepc`), and disabling compressed instructions altogether is not a good strategy either.

There are a few TODO items associated with this PR, mainly that it doesn't quite work with `NA4/NAPOT` modes yet due to an inconsistency in how addresses are passed in from the command line, this will be addressed in a separate PR.

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -22,11 +22,22 @@ class riscv_pmp_cfg extends uvm_object;
   // default to granularity of 0 (4 bytes grain)
   int pmp_granularity = 0;
 
+  // number of configuration bytes per pmpcfg CSR
+  int cfg_per_csr;
+
   // enable bit for pmp randomization
   bit pmp_randomize = 0;
 
   // allow pmp randomization to cause address range overlap
-  bit pmp_allow_addr_overlap = 0;
+  rand bit pmp_allow_addr_overlap = 0;
+
+  // By default, after returning from a PMP exception, we return to the exact same instruction that
+  // resulted in a PMP exception to begin with, creating an infinite loop of taking an exception.
+  // To avoid this situation, this configuration knob will enable the relevant PMP exception
+  // handlers to find the pmpcfg CSR that controls the address region resulting in the exception and
+  // change the relevant access bit to 1'b1, allowing forward progress in the code, while also
+  // allowing all access restrictions to be enforced.
+  bit enable_pmp_exception_handler = 1'b1;
 
   // pmp CSR configurations
   rand pmp_cfg_reg_t pmp_cfg[];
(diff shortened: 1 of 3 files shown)


src/riscv_pmp_cfg.sv
```diff
@@ -280,4 +304,279 @@ class riscv_pmp_cfg extends uvm_object;
     end
   endfunction
 
+  // This function creates a special PMP exception routine that is generated within the
+  // instr_fault, load_fault, and store_fault exception routines to prevent infinite loops.
+  // This routine will first find the correct pmpcfg CSR that corresponds to the address that
+  // caused the exception in the first place, and then will enable the appropriate access bit
+  // (X for instruction fauts, W for store faults, and R for load faults).
```
taoliug, line 311: fauts -> faults

udinator: good catch, thanks!

udinator:
> Please run this change with incisive to make sure the regression is passing
> 
> Sun, 26 Apr 2020 20:08:39 INFO parse_pmp_config(pmp_region, pmp_cfg[i]);
> |
> ncvlog: *E,UNSRFA (xxx/riscv_pmp_cfg.sv,174|43): unsupported ref argument usage because actual argument is not a supported element reference for ref arg.

That is actually an issue with IUS not having support for passing array elements by reference, though it is perfectly valid as per the LRM (we actually removed IUS support from Ibex environment for lack of support of many RTL/DV language features like this).
Should I restructure this to avoid this error?

taoliug:
> > Please run this change with incisive to make sure the regression is passing
> > Sun, 26 Apr 2020 20:08:39 INFO parse_pmp_config(pmp_region, pmp_cfg[i]);
> > |
> > ncvlog: *E,UNSRFA (xxx/riscv_pmp_cfg.sv,174|43): unsupported ref argument usage because actual argument is not a supported element reference for ref arg.
> 
> That is actually an issue with IUS not having support for passing array elements by reference, though it is perfectly valid as per the LRM (we actually removed IUS support from Ibex environment for lack of support of many RTL/DV language features like this).
> Should I restructure this to avoid this error?

Yes, I think we should make it work for the mainstream simulators

udinator:
> Yes, I think we should make it work for the mainstream simulators

Got it, will prepare a workaround.

udinator:
This latest force push contains a workaround to the issue of accessing structs in an array with the `ref` keyword.
@taoliug - I'm unable to reproduce the error you're seeing, could you see if this workaround is successful with IUS now?

udinator:
This force push takes into consideration a guideline in the spec that the combination of [W:1,R:0] is reserved for future use and should not be enabled.

udinator:
Latest force push fixes some merge conflicts introduced by #554.

PR #569, opened 2020-05-12
Restructure coverage

weicaiyang:
Change the structure of coverage for more extendability
Move coverage functions into riscv_instr and change most of functions to
virtual
Commented out most of floating point related functions for now, will move
them to riscv_floating_point_instr in another PR


src/isa/riscv_instr__cov.sv
```diff
@@ -0,0 +1,510 @@
...
+  special_val_e         rs2_special_val;
+  special_val_e         rs3_special_val;
+  special_val_e         rd_special_val;
+  special_val_e         imm_special_val;
+  compare_result_e      compare_result;
+  logical_similarity_e  logical_similarity;
+  string                trace;
+
+  // TODO, remove it?
+  //`VECTOR_INCLUDE("riscv_instr_cov_item_inc_declares.sv")
```
weicaiyang, line 66: most of codes in this file are copied from riscv_instr_cov_item.sv 
@taoliug somehow, I can't find `riscv_instr_cov_item_inc_declares.sv. do we still need it?

src/isa/riscv_instr.sv
```diff
@@ -649,4 +649,6 @@ class riscv_instr extends uvm_object;
     imm_str = $sformatf("%0d", $signed(imm));
   endfunction
 
+  `include "isa/riscv_instr__cov.sv"
```
taoliug, line 652: why use double underscore in the file name? Can there be a better name for this?

weicaiyang: it is to indicate that as an included file for riscv_instr? or we can name it as `riscv_instr_cov.svh` since we name some included file with `.svh`? WDYT?

taoliug: riscv_instr_cov.svh sgtm

weicaiyang: Updated, PTAL

PR #570, opened 2020-05-12
Add --seed_start argument and tidy up seed handling

rswarbrick:
*Note: I've edited this summary to match the version pushed up as `5dbfc25`.*

The `--seed_start` argument allows you to say "Please use the following
seed for the first iteration of each test. If you run more iterations
than one, increment the seed accordingly".

The idea is that some top-level program (for example Ibex's core_ibex
UVM test code) can pick a seed, possibly randomly, and then run many
tests using the same seeds for test generation (here) and test
execution (passed to the simulator that runs them).

We could implement that by generating a YAML file each time and
passing it with `--seed_yaml`, but that seems unnecessarily complicated.

While touching the seed handling code, I've also added checks for
various silly combinations of arguments that used to be possible.

  - If you specify `--seed` and `--iterations 10`, you'll now get an
    error. This is probably sensible because you're asking for 10
    identical tests, which is unlikely to be what you want.

  - If you specify `--seed` and don't specify `--iterations`, the
    iteration count is now forced to 1 rather than being read from the
    configuration file. Same reason as above.

The patch also pulls seed handling into a wrapper class (`SeedGen`),
which avoids having to pass yet another argument to `do_simulate`.

It also folds config file parsing into the end of `parse_args` and gets
rid of the call to `var(..)` at the end of `load_config`: we used to
convert from the `args` namespace to a dictionary called `cfg`, and then
promptly converted it back into a namespace. This seems a little
silly, so the patch gets rid of it.

Note: Before and after this patch, test seeds depend on the batch
size. This is because we don't re-seed between tests in a batch. For
example, these two commands will generate different programs:

    ./run.py --iterations=100 --batch-size 5 --seed_start 123
    ./run.py --iterations=100 --batch-size 10 --seed_start 123

It's not *too* bad, because the "test_id" which we use as a key in the
stored list of seeds is actually "test_name + batch index". But you'll
only be able to reproduce a generated test with a high iteration
number if you keep the batch size equal to the last time you ran.

This is easy enough to fix by passing an initial seed as a
plusarg (rather than using -svseed / -sv_seed arguments) and then
reseeding with, say, <seed> + <iteration> between each generated test.
I'll leave this for a later patch.

### DIFF
--- a/scripts/lib.py
+++ b/scripts/lib.py
@@ -85,20 +85,6 @@ def get_env_var(var, debug_cmd = None):
   return val
 
 
-def get_seed(seed):
-  """Get the seed to run the generator
-
-  Args:
-    seed : input seed
-
-  Returns:
-    seed to run instruction generator
-  """
-  if seed >= 0:
-    return seed
-  return random.getrandbits(31)
-
-
 def run_cmd(cmd, timeout_s = 999, exit_on_error = 1, check_return_code = True, debug_cmd = None):
   """Run a command and return output
 
--- a/run.py
+++ b/run.py
@@ -18,6 +18,7 @@ Regression script for RISC-V random instruction generator
 
 import argparse
 import os
+import random
 import re
 import sys
 import logging
@@ -32,6 +33,39 @@ from scripts.instr_trace_compare import *
 from types import SimpleNamespace
 
 LOGGER = logging.getLogger()
+
+
+class SeedGen:
+  '''An object that will generate a pseudo-random seed for test iterations'''
+  def __init__(self, start_seed, fixed_seed, seed_yaml):
+    # These checks are performed with proper error messages at argument parsing
+    # time, but it can't hurt to do a belt-and-braces check here too.
+    assert fixed_seed is None or start_seed is None
+
+    self.fixed_seed = fixed_seed
+    self.start_seed = start_seed
+    self.rerun_seed = {} if seed_yaml is None else read_yaml(seed_yaml)
+
+  def get(self, test_id, test_iter):
+    '''Get the seed to use for the given test and iteration'''
+
+    if test_id in self.rerun_seed:
+      # Note that test_id includes the iteration index (well, the batch index,
+      # at any rate), so this makes sense even if test_iter > 0.
+      return self.rerun_seed[test_id]
+
+    if self.fixed_seed is not None:
+      # Checked at argument parsing time
+      assert test_iter == 0
+      return self.fixed_seed
+
+    if self.start_seed is not None:
+      return self.start_seed + test_iter
+
+    # If the user didn't specify seeds in some way, we generate a random seed
+    # every time
+    return random.getrandbits(31)
+
 
 def get_generator_cmd(simulator, simulator_yaml, cov, exp, debug_cmd):
   """ Setup the compile and simulation command for the generator
@@ -186,7 +220,7 @@ def run_csr_test(cmd_list, cwd, csr_file, isa, iterations, lsf_cmd,
     run_cmd(cmd, timeout_s, debug_cmd = debug_cmd)
 
 
-def do_simulate(sim_cmd, test_list, cwd, sim_opts, seed_yaml, seed, csr_file,
+def do_simulate(sim_cmd, test_list, cwd, sim_opts, seed_gen, csr_file,
                 isa, end_signature_addr, lsf_cmd, timeout_s, log_suffix,
                 batch_size, output_dir, verbose, check_return_code, debug_cmd):
   """Run  the instruction generator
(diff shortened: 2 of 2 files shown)


rswarbrick:
Force-push fixes silly typo (return args.seed rather than args in parse_args).

taoliug:
The main purpose of the --seed option is to reproduce a simulation. If you want to run a larger regression, I think it's better to just use random seeds. There's also a way to reproduce the whole regression, I feel this is enough for most use cases. Please let me know your thought about this

rswarbrick:
How do you propose I run 100 tests in a nightly regression in a deterministic way? Also, see the Ibex issue above.

taoliug:
If you want to run the same seeds every night, you can use the below command:
python3 run.py -i 1 --seed_yaml out_xxx/seed.yaml

Every regression will generate seed.yaml under output directory, you can use it to run every night. However, from constraint random verification perspective, I think it's better to run with random seeds every night to uncover more issues. Fixing the seed every night may limit the state space that the test can explore.

rswarbrick:
Hmm, I'm not convinced. At the moment, using both `--seed S` and `--iterations N` for N > 1 is silly: you just end up generating the same test again and again. This patch makes the tool do something sensible (pick seeds starting at S and count up).

If you have a strong philosophical objection to this change (which I don't think I understand), the other sensible option would be to exit with an error when the user does this. The current behaviour silently does less testing than the user expected, which is definitely not good and is why @udinator opened the linked Ibex issue.

The difference between fixing the seeds every night, and searching new random seeds every night is the difference between regression testing and soak testing. At the moment, the tool supports soak testing by default and supports regression testing if I manually generate a YAML file for it (a bit of a faff). Since it's very simple to do so, I think we should also support regression testing by specifying a block of seeds. That's what this patch does.

udinator:
@rswarbrick I'll use this thread to respond to both this comment and the comment you left on the lowRISC/ibex#861,  since they're fairly similar.

I think one issue here that might be causing some confusion is that the `--seed S` option of the `run.py` script here and Ibex's `sim.py` are not really meant to be used in a "normal" use-case (e.g. running a single iteration of a test, running the full test suite, etc...), as both of these scripts will already generate a random seed for each iteration by default.
We only use `--seed S` if we see failure in some tests (say instruction generation fails for `riscv_dret_test_0` and RTL simulation fails for `riscv_ebreak_test_2`) and want to reproduce these failures (in this case we could do some debugging and then rerun the generation phase of `riscv_dret_test_0` and RTL simulation phase of `riscv_ebreak_test_2`, both with their respective `--seed S` options specified).
As a result, explicitly specifying the combination of `--seed S` and `--iterations N` where N>1 is a situation that generally shouldn't be called by the user (but is what the current Ibex Makefile does for both scripts) due to this fact that these scripts should already be generating seeds for each iteration.
I think this applies to the Ibex makefile as well, the `SEED` variable should only be defined by the user when trying to reproduce a particular failing sim, not for general purpose testing.

With regards to your comments about regression vs. soak testing - we use different seeds for the OpenTitan IP nightly regressions instead of repeating seeds each night. A main reason for this is that if some IP is already passing 100% of its tests, running that same set of seeded tests each night doesn't provide as high a level of confidence in the correctness of our tests/designs as does running completely different seeds every night and seeing 100% pass rates every night. There is also the distinct possibility that debugging regression errors and then only running that same set of seeds every night will mask other hidden bugs, which makes reproducing an entire regression useful mostly just for initial debugging purposes.

Currently, riscv-dv automatically generates a yaml file (`seed.yaml`) containing mappings between every test iteration and the seed of that iteration (no manual work is needed to create this file); this is currently auto-generated at `<out>/seed-#/instr_gen/seed.yaml` when running the Ibex makefile. All that is needed to reproduce regression results with the same set of seeds is simply to add 
```console
GEN_OPTS="--seed_yaml <out>/seed-#/instr_gen/seed.yaml"
```
the next time we run the Makefile and the generation flow will pick up and use the exact same set of seeds, which I believe addresses your concerns about reproducibility of regressions.

The RTL side of the Ibex flow does not support creating this seed.yaml file to natively be able to reproduce regressions but will be an easy feature to add, which is what I was alluding to in my comment on the Ibex PR; I think this approach of letting the scripts automatically randomize every seed and reserving use of `--seed S` just to reproduce simulations (this was the original approach we used for Ibex sims) instead of being the mechanism through which the Ibex Makefile passes top level seeds to the scripts (which then causes the scripts to have extra logic just to figure out if they should increment the seeds or not, etc...) is a bit cleaner.

tomroberts-lowrisc:
If I understand it correctly, this change doesn't have any effect on the normal use-cases of re-running a single failing test or running a suite of random seeds, it just makes the behavior more useful if the user specifies multiple tests with a particular seed.

Given that, I think this is a reasonable enhancement. Other test running frameworks I have used in the past have had this feature (in addition to specifying a list of seeds in a file) for smoke testing etc.

taoliug:
I think it's nice to add this support with a separate option rather than using the existing --seed which is meant to be used as a fixed seed to reproduce a simulation.

rswarbrick:
Ok, if you think that would be better. I've just pushed up a new version of the patch that adds a `--start_seed` parameter to do this. The commit message has changed somewhat, so I'll also update the message at the top of the PR to match.

scripts/lib.py
```diff
@@ -85,20 +85,6 @@ def get_env_var(var, debug_cmd = None):
   return val
 
 
-def get_seed(seed):
```
udinator, line 88: note that Ibex sim.py uses this function, please make sure to update the Ibex script to match what is happening here.

rswarbrick: Good point. I'll add it to a vendoring patch today.

PR #579, opened 2020-05-19
Add Xcelium support

tudortimi:
No coverage (yet).


tudortimi:
I want to get a test suite working in Ibex before I add coverage.

tudortimi:
@googlebot I signed it!

PR #580, opened 2020-05-19
convert handshake doc to rst format

udinator:


udinator:
> Can you generate the doc in simplehtml format?

Done.

udinator:
Hmm, I wasn't sure if the upstream documentation needed these binaries to be checked in or not.
Let me remove them from the PR and re-push.

PR #600, opened 2020-06-08
Add floating point coverage part2

weicaiyang:
for #443, including all remaining items
1. 11.7-11.9
2. chapter 12 double precision
3. floating point special values

### DIFF
--- a/src/riscv_instr_cover_group.sv
+++ b/src/riscv_instr_cover_group.sv
@@ -228,17 +228,54 @@
   `INSTR_CG_BEGIN(INSTR_NAME) \
     cp_imm_sign : coverpoint instr.imm_sign;
 
-// TODO, will handle special value later
-// single-precision floating point special values coverpoint
-`define FP_SPECIAL_VALUES_CP(VAR, NAME) \
-    cp_fp_special_values_on_``NAME`` : coverpoint VAR { \
+// single/double precision floating point special values coverpoint
+`define FP_SPECIAL_VALUES_CP(VAR, NAME, PRECISION = S) \
+    cp_sfp_special_values_on_``NAME`` : coverpoint VAR { \
+      option.weight = (`"PRECISION`" == "S"); \
+      type_option.weight = (`"PRECISION`" == "S"); \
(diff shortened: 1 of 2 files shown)


src/riscv_instr_cover_group.sv
```diff
@@ -228,27 +228,67 @@
...
+// single/double precision floating point special values coverpoint
+`define FP_SPECIAL_VALUES_CP(VAR, NAME, PRECISION = S) \
+    cp_sfp_special_values_on_``NAME`` : coverpoint VAR { \
+      option.weight = (`"PRECISION`" == "S"); \
+      type_option.weight = (`"PRECISION`" == "S"); \
       bins infinity[] = {32'h7f80_0000, 32'hff80_0000}; \
       bins largest[]  = {32'h7f7f_ffff, 32'hff7f_ffff}; \
-      bins zeros[]    = {32'h0000_0000, 32'h1000_0000}; \
+      bins zeros[]    = {32'h0000_0000, 32'h8000_0000}; \
       bins NaN[]      = {32'h7fc0_0000, 32'h7f80_0000}; \
```
taoliug, line 239: is 7f80_0000 Nan?

weicaiyang: in riscv spec, 7fc0_0000 is quite qNaN, so flipping the MSB significand bit (quite bit) should become sNaN which is 7f80_0000. however, h7f80_0000 is a also positive infinity value according to [wiki](https://en.wikipedia.org/wiki/Single-precision_floating-point_format)

not sure if I should use 32'h7fc0_000**1,** 32'h7f80_000**1** as NaN values. Will add direct test to try on FCLASS as it will tell these kinds of values
```
rd bit Meaning
0 rs1 is −∞.
1 rs1 is a negative normal number.
2 rs1 is a negative subnormal number.
3 rs1 is −0.
4 rs1 is +0.
5 rs1 is a positive subnormal number.
6 rs1 is a positive normal number.
7 rs1 is +∞.
8 rs1 is a signaling NaN.
9 rs1 is a quiet NaN
```

PR #625, opened 2020-06-25
Importing PyVSC module

hodjat91:
Just imported pyvsc into riscv_instr.py and riscv_instr_gen_config.py.


taoliug:
Should we add this package to https://github.com/google/riscv-dv/blob/master/requirements.txt ?

hodjat91:
> Should we add this package to https://github.com/google/riscv-dv/blob/master/requirements.txt ?
Yeah makes sense

Issue #603, opened 2020-06-10
Initialization of vector registers don't scale well


Sustrak:
With the new implementation of memory operations for the vector extension the initialization of the vector registers has changed.

Before the update, vector registers were initialized using the instruction `vmv.v.x` and so all the elements in the vector register had the same value. With the new update each element is generated randomly using the instructions `vmv.v.x` and `vslide1up`.

One problem with this is when generating code for a vector machine with a big VLEN, for example in our case VLEN = 16384 (256x64) the code for initilizing the vector registers is +20k LOC long.

It would be nice to have different configurations for initilizing the vector registers I think of 3 of them:

1. Using only `vmv.v.x`.
2. Using `vmv.v.x` and `vslide1up`.
3. Using `vle.v`.

For vle I implemented this in the `init_vec_gpr` function:
```
for (int v = 1; v < NUM_VEC_GPR; v++) begin
      int region = $urandom_range(0, 4);
      instr_stream.push_back($sformatf("%0sla t0, region_%0d", indent, region));
      instr_stream.push_back($sformatf("%0svle.v v%0d, (t0)", indent, v));
end
```

taoliug:
SGTM, I will add support for these three modes

taoliug:
Fixed https://github.com/google/riscv-dv/pull/627

PR #627, opened 2020-06-30
Add different methods to initialize the vregs

Sustrak:
Three different methods have been implemented to initialize the vector registers.
We have mantained the current one that uses the instructions vmv and vslide1up and added
the method that was implemented before that used only the vmv instruction and implemented
a new method that uses the vle instruction.

This feature has been discussed in the issue #603.

This work has been done by the verification team in the EPI project at BSC-CNS.

### DIFF
--- a/target/ml/riscv_core_setting.sv
+++ b/target/ml/riscv_core_setting.sv
@@ -80,12 +80,6 @@ parameter int VELEN = int'($ln(ELEN)/$ln(2)) - 3;
 // Maxium LMUL supported by the core
 parameter int MAX_LMUL = 8;
 
-// Initialization of the vregs
-// SAME_VALUES_ALL_ELEMS - Using vmv.v.x to fill the all the elements of the vreg with the same value as the GPR selected
-// RANDOM_VALUES_VMV - Using vmv.v.x + vslide1up.vx to randomize the contents of each vector element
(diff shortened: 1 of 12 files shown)


Sustrak:
@googlebot I fixed it.

target/ml/riscv_core_setting.sv
```diff
@@ -80,6 +80,12 @@ parameter int VELEN = int'($ln(ELEN)/$ln(2)) - 3;
 // Maxium LMUL supported by the core
 parameter int MAX_LMUL = 8;
 
+// Initialization of the vregs
+// SAME_VALUES_ALL_ELEMS - Using vmv.v.x to fill the all the elements of the vreg with the same value as the GPR selected
+// RANDOM_VALUES_VMV - Using vmv.v.x + vslide1up.vx to randomize the contents of each vector element
+// RANDOM_VALUES_LOAD - Using vle.v, same approach as RANDOM_VALUES_VMV but more efficient for big VLEN
+parameter vreg_init_method_t VREG_INIT_METHOD = RANDOM_VALUES_VMV;
```
taoliug, line 87: Can you add this to riscv_instr_gen_config rather than core_setting.sv? This knob is not a processor feature, it's a configuration of how to initialize the vector GPR

Sustrak: Done, I'm not sure if it should be random so I didn't set it.

PR #646, opened 2020-07-15
added uvm_glob_to_re in uvm_re_match

dawidzim:
Hey!
I think we need to convert `*main*` to regular expressions by `uvm_glob_to_re`


dawidzim:
@googlebot I signed it!

PR #653, opened 2020-07-20
Add functional coverage test file 

hodjat91:
Add functional coverage test file.


pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -0,0 +1,136 @@
...
+                                if not expect_illegal_instr:
+                                    logging.error("Found unexpected illegal instr: {} [{}]".format(self.trace["instr"], entry))
+                                    self.unexpected_illegal_instr_cnt += 1
+                            # TODO: call sample here to do the functional coverage, and move assign_trace_info into it
+                            #self.sample(instruction)
+                        self.entry_cnt += 1
+                    line_count += 1
+                logging.info("[{}]: {} instruction processed".format(csv_file, self.entry_cnt))
+                self.total_entry_cnt += self.entry_cnt
+        logging.info("Finished processing {} trace CSV, {} instructions".format(len(self.trace_csv), self.total_entry_cnt))
```
udinator, line 84: nit: here and in a few other places, keep line length to 80

hodjat91: Thank Udi, I will fix it.

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -0,0 +1,136 @@
...
+                               "lr" in self.trace["instr"] or \
+                               "sc" in self.trace["instr"]:
+                                # TODO: Enable functional coverage for AMO test
+                                pass
+                            if not self.sample():
+                                if not expect_illegal_instr:
+                                    logging.error("Found unexpected illegal instr: {} [{}]".format(self.trace["instr"], entry))
+                                    self.unexpected_illegal_instr_cnt += 1
+                            # TODO: call sample here to do the functional coverage, and move assign_trace_info into it
+                            #self.sample(instruction)
```
udinator, line 79: Are these comments necessary, since `self.sample()` is already called above?

hodjat91: Oops, that's already called. I'll fix this one too.

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -0,0 +1,136 @@
...
+def parse_args():
+    parse = argparse.ArgumentParser()
+    parse.add_argument("--csv_dir", help="directory consists of the CSV files", type=str, default = ".")
+    args = parse.parse_args()
+    return args
+
+def main():
+    parse = argparse.ArgumentParser()
+    parse.add_argument("--csv_dir", help="directory consists of the CSV files", type=str, default = ".")
+    args = parse.parse_args()
```
udinator, line 131: do you mean to call the `parse_args()` defined above instead?

udinator: sorry, I take back this comment actually.

it's actually unnecessary for this coverage test to search for all the `.csv` files that are generated, as the `cov.py` script already gets the pathname to the CSV filenames [here](https://github.com/google/riscv-dv/blob/master/cov.py#L137-L147) and passes them all into the coverage test as cmdline arguments - you should be able to just iterate through `sys.argv` and pull the pathnames from there directly.

pygen/pygen_src/isa/riscv_instr_cov.py
```diff
@@ -0,0 +1,44 @@
...
+    MIN_VAL = auto()
+    MAX_VAL = auto()
+    ZERO_VAL = auto()
+
+# dictionary to keep track of reg_name:reg_value throughout the program 
+gpr_state = {}
+
+def get_gpr(reg_name):
+    reg_name = reg_name.upper()
+    # logging.error...
```
udinator, line 36: maybe a helper function is needed here to convert a string into the corresponding `Enum` member (in this case converting `reg_name` into `riscv_reg_t` Enum type)?
https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -0,0 +1,194 @@
...
+            logging.error("{} instruction skipped, {} illegal "
+                          "instructions".format(self.skipped_cnt),
+                          self.unexpected_illegal_instr_cnt)
+
+    def post_process_trace(self):
+        pass
+
+    def sample(self):
+        instr_name, binary = "", ""
+        get_val(self.trace["binary"], binary, hexa=1)
```
udinator, line 115: maybe we should have a standardized usage of `get_val(...)`? 
Seems like right now the usage depends on the type of variable it returns (e.g. here we pass the `binary` string by reference but below on line 133 we explicitly assigns the return value to `instruction.pc` - maybe we can eliminate the `output_val` argument of `get_val()` and always explicitly assign a return value for clarity?

pygen/pygen_src/isa/riscv_instr_cov.py
hodjat91: Thanks for the comment! Fixed.

pygen/pygen_src/test/riscv_instr_cov_test.py
hodjat91: That makes sense. It's fixed.

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
udinator, line 1: Several `.idea/*` files seem to have been added?

hodjat91: Oh this is the .idea directory created by the IDE, fixed it.

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -0,0 +1,293 @@
...
+
+class riscv_instr_cov_test(argv):
+    """ Main class for applying the functional coverage test """
+    def __init__(self):
+        self.trace = {}
+        self.entry_cnt, self.total_entry_cnt, self.skipped_cnt, \
+        self.unexpected_illegal_instr_cnt = 0, 0, 0, 0
+
+    def run_phase(self):
+        logging.info("{} CSV trace files to be "
```
udinator, line 162: should we short-circuit and `sys.exit` right away if argv is empty?

hodjat91: Yup, makes sense.

PR #654, opened 2020-07-22
[Pygen] Fix minor Issues

aneels3:
- Fix indentation Issue in generated assembly file

- Fix typo in extend_imm() function. 

- Fix imm_c constraint (Temporary Solution).

### DIFF
--- a/pygen/pygen_src/isa/riscv_instr.py
+++ b/pygen/pygen_src/isa/riscv_instr.py
@@ -19,7 +19,7 @@ import os
 import vsc
 from collections import defaultdict
 from bitstring import BitArray
-from pygen_src.riscv_instr_pkg import pkg_ins, riscv_reg_t
+from pygen_src.riscv_instr_pkg import pkg_ins, riscv_reg_t, riscv_instr_name_t
 from pygen_src.isa import rv32i_instr  # NOQA
 from pygen_src.target.rv32i import riscv_core_setting as rcs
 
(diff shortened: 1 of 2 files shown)


pygen/pygen_src/isa/riscv_instr.py
```diff
@@ -271,14 +271,31 @@ def extend_imm(self):
         sign = (self.imm & 0x80000000) >> 31
         self.imm = self.imm >> (32 - self.imm_len) & self.shift_t
         # Signed extension
-        if((sign and not(self.format == "U_FORMAT")) or (self.imm_type in ["UIMM", "NZUIMM"])):
+        if(sign and not((self.format.name == "U_FORMAT") or
+                        (self.imm_type.name in ["UIMM", "NZUIMM"]))):
             self.imm = self.imm_mask | self.imm
 
+    def imm_c(self):
```
hodjat91, line 278: Hi @aneels3
I understand that this is a temporary solution for the imm_c constraint (and it looks good to me), but my question is why do we need a temporary solution? I thought the issue with imm_c constraint via PyVSC is solved... Or is it due to a gradual transition to the PyVSC usage throughout the generator?

aneels3: Hi @hodjat91 
I forgot to mention that the the implementation with PyVSC is creating problem with the overall flow although the implementation is correct.

When we declare `self.instr_name = vsc.enum_t(riscv_instr_name_t)`
The generator is stuck at create_instr() function. It is not able to create the instances of instruction classes.

hodjat91: Hi @aneels3 
I see... Have you investigated on the design to see where does the issue come from? BTW, the `self.instr_name` should actually be `vsc.rand_enum_t(riscv_instr_name_t)`, right? Please let me know if I can be of further assistance with this regard (it actually worked fine in a simple version that I created (although is only generates arithmetic instructions with a few constraints)).

aneels3: Hi @hodjat91 
Yes, Absolutely correct.
Before while discussing on the issues section on this we thought the problem is resolved. Sorry about that!
I am also thinking to map it with PyVSC but after investigating a lot didn't able to resolve the problem.
Because of this constraint, the SLLI SRAI, etc. instructions are being randomized with an incorrect value, and the same was reflected while compiling the generated Assembly file.
So, for the time being, I am going with the temporary one.

Any suggestions on this?

taoliug: OK to move forward with temp workaround as long as we have a TODO to fix it later.

PR #662, opened 2020-07-28
Added methods to the coverage test file

hodjat91:
More methods will be added to complete the coverage test file.

### DIFF
--- a/pygen/pygen_src/isa/riscv_instr_cov.py
+++ b/pygen/pygen_src/isa/riscv_instr_cov.py
@@ -5,27 +5,27 @@ from enum import Enum, auto
 from pygen_src.riscv_instr_pkg import riscv_reg_t
 
 class operand_sign_e(Enum):
-    POSITIVE = auto()
+    POSITIVE = 0
     NEGATIVE = auto()
 
 class div_result_e(Enum):
-    DIV_NORMAL = auto()
+    DIV_NORMAL = 0
     DIV_BY_ZERO = auto()
     DIV_OVERFLOW = auto()
 
 class compare_result_e(Enum):
-    EQUAL = auto()
+    EQUAL = 0
     LARGER = auto()
     SMALLER = auto()
 
 class logical_similarity_e(Enum):
-    IDENTICAL = auto()
+    IDENTICAL = 0
     OPPOSITE = auto()
     SIMILAR = auto()
     DIFFERENT = auto()
 
 class special_val_e(Enum):
-    NORMAL_VAL = auto()
+    NORMAL_VAL = 0
     MIN_VAL = auto()
     MAX_VAL = auto()
     ZERO_VAL = auto()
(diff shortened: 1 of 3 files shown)


pygen/pygen_src/isa/riscv_instr_cov.py
```diff
@@ -5,27 +5,27 @@
 from pygen_src.riscv_instr_pkg import riscv_reg_t
 
 class operand_sign_e(Enum):
-    POSITIVE = auto()
+    POSITIVE = 0
```
udinator, line 8: why hardcode these to zero?

hodjat91: No specific reason actually, just wanted to have an auto enum starting from 0.

udinator: ah, I just checked the auto() documentation and seems like it starts from 1 by default - I'm fine starting these from 0.

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -105,9 +106,10 @@ def pre_sample(self):
                           "REMW", "REMUW"]:
             self.div_result = self.get_div_result()
 
-    def get_operand_sign(self, value):
-        #TODO: cast value to vsc.bit_t
-        if value[0]:
+    def get_operand_sign(self, operand):
+        #TODO: change operand to vsc.bit_t
```
udinator, line 110: Is there anything blocking this TODO from being implemented in this PR?

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -125,26 +127,134 @@ def is_unaligned_mem_access(self):
...
+                return special_val_e["MAX_VAL"]
+        return special_val_e["NORMAL_VAL"]
 
-    def get_logical_similarity(self):
-        pass
+    def get_compare_result(self):
+        val1 = vsc.int_t(rcs.XLEN)
+        val2 = vsc.int_t(rcs.XLEN)
+        val1.val = self.rs1_value.val
+        val2.val = self.imm.val if (
```
udinator, line 182: this is a pretty nitty comment, but  why not instantiate these as `val1 = vsc.int_t(rcs.XLEN, self.rs1_value.val)` to match what you do in other functions?

pygen/pygen_src/test/riscv_instr_cov_test.py
hodjat91: Nothing is blocking it actually. But as you can see, I have a bunch of these "change type" TODOs that need to be reviewed/done after completing the coverage test file (and its methods). Also some of these TODOs (those attributes belong to a `riscv_inst` object that are already defined in `riscv_instr.py` file) are dependent to the way they are defined in that file (which is not fully based on pyvsc yet).

pygen/pygen_src/test/riscv_instr_cov_test.py
hodjat91: Fixed 👍

pygen/pygen_src/test/riscv_instr_cov_test.py
udinator: got it, sgtm

PR #664, opened 2020-07-29
Add new PMP directed sequence

udinator:
This patch adds in a new directed sequence of instructions placed
directly after the PMP setup.
This directed sequence will write randomized values to every `pmpaddr`
and `pmpcfg` CSR in order to test write accessibility (and spec
compliance). Correctness checking for this sequence is determined by
post-comparison with the ISS log, as per usual.

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -39,6 +39,10 @@ class riscv_pmp_cfg extends uvm_object;
   // allowing all access restrictions to be enforced.
   bit enable_pmp_exception_handler = 1'b1;
 
+  // Setting this bit to 1'b1 enables generation of the directed stream of instructions to test
+  // write accesses to all supported pmpaddr[i] CSRs.
+  bit enable_write_pmp_csr;
+
   // pmp CSR configurations
   rand pmp_cfg_reg_t pmp_cfg[];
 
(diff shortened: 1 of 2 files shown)


src/riscv_pmp_cfg.sv
```diff
@@ -606,4 +611,54 @@ class riscv_pmp_cfg extends uvm_object;
...
+
+    bit [11:0] pmp_addr;
+    bit [11:0] pmpcfg_addr;
+    bit [XLEN-1:0] pmp_val;
+    for (int i = 0; i < pmp_num_regions; i++) begin
+      pmp_addr = base_pmp_addr + i;
+      pmpcfg_addr = base_pmpcfg_addr + (i / cfg_per_csr);
+      // Keep all randomized addresses well over the "safe" region to avoid
+      // corrupting the initialization routine.
+      `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(pmp_val, pmp_val >= 'h8000_4000;)
```
taoliug, line 627: Can 8000_4000 be a parameter? User may link the program to different location.

udinator: Sure, I can look into having some sort of parameter for this value

src/riscv_pmp_cfg.sv
```diff
@@ -606,4 +611,54 @@ class riscv_pmp_cfg extends uvm_object;
...
+      instr.push_back($sformatf("csrrw x%0d, 0x%0x, x%0d", scratch_reg, pmp_addr, scratch_reg));
+      // Restore the original address to pmpaddr[i].
+      // New value of pmpaddr[i] will be written to scratch_reg.
+      instr.push_back($sformatf("csrrw x%0d, 0x%0x, x%0d", scratch_reg, pmp_addr, scratch_reg));
+      // Randomize value to be written to pmpcfg CSR.
+      //
+      // TODO: support rv128?
+      `DV_CHECK_STD_RANDOMIZE_WITH_FATAL(pmp_val,
+                                         // Ensure that the random config regions aren't locked
+                                         if (XLEN == 64) {
```
udinator, line 640: oops, just realized that this constraint is actually incorrect and needs to be redone, will push a new version in the next force push

src/riscv_pmp_cfg.sv
udinator: I think a better option might be to just ensure that `pmp_val` is greater than the address of `<main>`, this way we keep any necessary relative addressing, while accomplishing the same end result.

udinator:
latest force push fixes how we randomize values written to `pmpaddr[i]` and fixes the randomization issue to explicitly zero every 8th bit of the value written to `pmpcfg[i]`

PR #667, opened 2020-07-30
Add gpr_c constraint

aneels3:
Just adding the constraint for GPR.

`not_inside` is not yet supported in PyVSC when we declare a list of enum types of fix size as `vsc.rand_list_t(vsc.enum_t(riscv_reg_t), sz =4)`


saurabhsingh-pvips:
@udinator @hodjat91 
FYI, you can see the coverage of instructions till now here: https://github.com/saurabhsingh-pvips/riscv-dv/tree/coverage_report.
urgReport/dashboard.html will display the coverage summary.

PR #665, opened 2020-07-29
[Pygen]: Modify asm_program functions.

aneels3:
Add more functions of the **riscv_asm_gen_program.py** file to achieve the compilation and simulation of the generated assembly file.
Achieve simulation of the current assembly file in both SPIKE and OVPsim.

PR #672, opened 2020-08-06
[pygen]: Add post_process and ecall_handler

aneels3:
Update the **post_process** function of **riscv_instr_sequence.py** for branch instructions.
Added **ecall_handler** which was missed in #665

### DIFF
--- a/pygen/pygen_src/riscv_asm_program_gen.py
+++ b/pygen/pygen_src/riscv_asm_program_gen.py
@@ -17,7 +17,7 @@ import logging
 import random
 from bitstring import BitArray
 from pygen_src.riscv_instr_sequence import riscv_instr_sequence
-from pygen_src.riscv_instr_pkg import (pkg_ins, privileged_reg_t)
+from pygen_src.riscv_instr_pkg import pkg_ins, privileged_reg_t, privileged_mode_t, mtvec_mode_t
 from pygen_src.riscv_instr_gen_config import cfg
 from pygen_src.target.rv32i import riscv_core_setting as rcs
 '''
@@ -103,9 +103,31 @@ class riscv_asm_program_gen:
             logging.info("Main/sub program generation...done")
             # program end
             self.gen_program_end(hart)
+            for hart in range(cfg.num_of_harts):
+                self.gen_data_page_begin(hart)
+                if(cfg.no_data_page):
+                    self.gen_data_page()
+
+                    if((hart == 0) and ("RV32A" in rcs.supported_isa)):
+                        self.gen_data_page(hart, amo = 1)
+
+            self.gen_stack_section(hart)
+            if(not cfg.bare_program_mode):
+                self.gen_kernel_sections(hart)
 
     def gen_kernel_sections(self, hart):
-        pass
+        if(rcs.SATP_MODE != "BARE"):
+            self.instr_stream.append(".align 12")
+        else:
+            self.instr_stream.append(".align 2")
+        self.instr_stream.append(pkg_ins.get_label("kernel_instr_start:", hart))
+        self.instr_stream.append(".text")
+
+        self.gen_all_trap_handler(hart)
+        for mode in rcs.supported_privileged_mode:
+            self.gen_interrupt_handler_section(mode, hart)
+
+        self.gen_kernel_stack_section(hart)
 
     def gen_kernel_program(self, hart, seq):
         pass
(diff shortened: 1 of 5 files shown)


pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -292,7 +293,20 @@ def gen_test_done(self):
...
-        pass
+        string = ""
+        # load base address
+        string = "{} la x{}, _start".format(pkg_ins.indent, cfg.gpr[0].value)
+        self.instr_stream.append(string)
+
+        # Generate sw/sd instructions
+        for i in range(32):
+            if (rcs.XLEN == 64):
+                string = "{} sd x{}, {}(x{})".format(
```
udinator, line 304: nit here and below: `{} sd` -> `{}sd`

pygen/pygen_src/riscv_instr_sequence.py
```diff
@@ -1,89 +1,191 @@
...
+                    not(self.instr_stream.instr_list[i].atomic)):
+                if((self.illegal_instr_pct > 0) and
+                   (self.instr_stream.instr_list[i].insert_illegal_instr == 0)):
+                    '''
+                    The illegal instruction generator always increase PC by 4 when resume execution,
+                    need to make sure PC + 4 is at the correct instruction boundary.
+                    '''
+                    if(self.instr_stream.instr_list[i].is_compressed):
+                        if(i < (len(self.instr_stream.instr_list) - 1)):
+                            if(self.instr_stream.instr_list[i + 1].is_compressed):
```
udinator, line 102: nit: `[i + 1]` -> `[i+1]`

pygen/pygen_src/riscv_instr_sequence.py
```diff
@@ -1,89 +1,191 @@
...
+                                self.instr_stream.instr_list[i].is_illegal_instr = random.randrange(
+                                    0, min(100, self.illegal_instr_pct))
+                        else:
```
udinator, line 105: the `else` statement is aligned with the wrong `if` statement, this should be aligned with the `if(self.instr_stream.instr_list[i].is_compressed)`

pygen/pygen_src/riscv_asm_program_gen.py
aneels3: Thanks!
Fixed now.

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -292,7 +293,20 @@ def gen_test_done(self):
...
+        string = ""
+        # load base address
+        string = "{}la x{}, _start".format(pkg_ins.indent, cfg.gpr[0].value)
+        self.instr_stream.append(string)
+
+        # Generate sw/sd instructions
+        for i in range(32):
+            if (rcs.XLEN == 64):
+                string = "{}sd x{}, {}(x{})".format(
+                    pkg_ins.indent, i, i * (rcs.XLEN / 8), cfg.gpr[0].value)
```
hodjat91, line 305: Very minor, but could you please change rcs.XLEN / 8 to rcs.XLEN // 8?

pygen/pygen_src/riscv_instr_sequence.py
```diff
@@ -1,89 +1,191 @@
...
+    lead to dead loops in the absence of proper loop exiting conditions.
+    ----------------------------------------------------------------------------------------------
+    '''
+
+    def post_process_instr(self):
+        label_idx = 0
+        branch_cnt = 0
+        j = 0
+
+        def def_val():
```
hodjat91, line 80: Can't we take this `def_val` function out of the `post_process_instr`? What's the purpose of it?

aneels3: Hi @hodjat91 
I have taken branch_target as `defaultdict()` such that it doesn't throw any key error.
So, to initialize the dictionary with a default value i.e 0, I have taken the def_val() function because the first argument in the defaultdict() must be **callable** or **None**. 

I can think of 2 scenarios here
1) Like you said to define the def_val() function outside the post_process() function. or
2) Provide the default value as None instead of 0.

For the second case, we wont need the def_val() function.
What do you think?

hodjat91: Thanks for the clarification, @aneels3 . I got it. I think either way should be fine, though I prefer the second option as it avoids using helper function (which might make it a bit confusing).

aneels3: Fixed Now!
Thanks @hodjat91

PR #680, opened 2020-08-17
Coverage report visualization supported + Few covergroups are added to complete RV32I CGs

hodjat91:
This PR:
- Adds support for coverage report visualization (by storing the coverage results in XML format to be read by pyucis-viewer)
- Adds CSR, opcode, rv32i_misc, and mepc_alignment covergroups

### DIFF
--- a/pygen/pygen_src/riscv_instr_pkg.py
+++ b/pygen/pygen_src/riscv_instr_pkg.py
@@ -24,7 +24,7 @@ class mem_region_t:
     xwr = auto()
 
 
-class satp_mode_t(Enum):
+class satp_mode_t(IntEnum):
     BARE = 0b0000
     SV32 = 0b0001
     SV39 = 0b1000
@@ -33,7 +33,7 @@ class satp_mode_t(Enum):
     SV64 = 0b1011
 
 
-class f_rounding_mode_t(Enum):
+class f_rounding_mode_t(IntEnum):
     RNE = 0b000
     RTZ = 0b001
     RDN = 0b010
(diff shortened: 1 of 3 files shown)


pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -116,6 +116,8 @@ def get_coverage_report():
         file.write(tabulate(table, headers, tablefmt="grid",
                             numalign="center", stralign="center"))
         file.close()
+        # Write in xml format to be read by pyucis-viewer (visulization)
```
udinator, line 119: nit: `visulization` -> `visualization`

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -116,6 +116,8 @@ def get_coverage_report():
...
+        vsc.write_coverage_db("cov_db.xml")
```
udinator, line 120: just to confirm, after we save this `cov_db.xml`, is the user able to open this with UCIS-viewer from the command line?

pygen/pygen_src/riscv_instr_pkg.py
```diff
@@ -24,7 +24,7 @@ class mem_region_t:
     xwr = auto()
 
 
-class satp_mode_t(Enum):
+class satp_mode_t(IntEnum):
```
udinator, line 27: it doesn't seem like it, but switching everything over to IntEnum will not affect anything throughout the generator, correct?

pygen/pygen_src/test/riscv_instr_cov_test.py
hodjat91: That is correct, yes. Basically what happens is that we add an additional command line option to the `cov.py` file for visualization purpose: if the flag is enabled, then this `cov_db.xml` will be generated, then as you said, the user will be able to read this xml file with UCIS-Viewer from the command line and see the output (alternatively, we can embed this functionality as well but in that case we need to add the `pyucis-viewer` package to the `requirements.txt` file, and also include the source folder of `pyucis-viewer` in the `riscv-dv/pygen` directory, which might not be a good idea at the moment. With that said, I suggest to leave it for user to read the file using pyucis-viewer). What do you think?

pygen/pygen_src/riscv_instr_pkg.py
hodjat91: No it doesn't affect , and I've also confirmed this with @aneels3 that switching to `IntEnum` is a don't care for their implementation.

udinator: sounds good to me!

pygen/pygen_src/test/riscv_instr_cov_test.py
udinator: got it. I think this is fine as is for right now - if we ever get to the point where we are comfortable integrating `pyucis-viewer` into the end-to-end run flow, we can add functionality at that later time.

hodjat91: Makes sense, sure :-)

PR #684, opened 2020-08-20
Remove f strings for Python 3.5-compatibility

imphil:
RISC-V DV is mostly Python 3.5-compatible, with the exception of the use
of f strings in gen_csr_test.py. Replace those with Python
3.5-compatible `.format()` formatting.


imphil:
OpenTitan/Ibex have a Python 3.5 requirement, which motivated this change.

imphil:
@udinator can you merge for me?

PR #688, opened 2020-08-27
[Pygen]: Add jal instr

aneels3:
PR for riscv_jal_instr!


ShraddhaDevaiya:
@googlebot I consent.

saurabhsingh-pvips:
@googlebot I consent.

pvipsyash:
@googlebot I consent.

pygen/pygen_src/isa/riscv_instr.py
```diff
@@ -188,7 +188,10 @@ def get_rand_instr(self, include_instr=[], exclude_instr=[],
         if(len(disallowed_instr) == 0):
             try:
                 if(len(include_instr) > 0):
-                    idx = random.randrange(0, len(include_instr) - 1)
+                    if(len(include_instr) == 1):
```
hodjat91, line 191: nit: Please remove the `()` in `if` statements like this (with a single condition): according to PEP8, parentheses in `if` statements are redundant.

pygen/pygen_src/riscv_directed_instr_lib.py
```diff
@@ -11,34 +11,99 @@
...
+        order = []
+        RA = cfg.ra
+        order = [0] * self.num_of_jump_instr
+        self.jump = [0] * self.num_of_jump_instr
+        for i in range(len(order)):
+            order[i] = i
+        random.shuffle(order)
+        self.setup_allowed_instr(1, 1)
+        jal = [riscv_instr_name_t.JAL]
+        if(not cfg.disable_compressed_instr):
```
hodjat91, line 68: Same here, redundant parentheses.

pygen/pygen_src/isa/riscv_instr.py
aneels3: Fixed!

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -629,7 +629,7 @@ def generate_directed_instr_stream(self, hart = 0, label = "", original_instr_cn
                 if(object_h is None):
                     logging.critical("Cannot create instr stream %0s", name)
                     sys.exit(1)
-                new_instr_stream = copy.copy(object_h)
+                new_instr_stream = copy.deepcopy(object_h)
```
udinator, line 632: is this going to incur significant performance penalties for tests that use a lot of directed streams?

pygen/pygen_src/riscv_instr_gen_config.py
```diff
@@ -413,9 +413,14 @@ def parse_args():
...
-    parse.add_argument('--stream_name_opts', help = 'stream_name_0',
+    parse.add_argument('--directed_instr_6', help = 'directed_instr_6',
+                       default = "riscv_jal_instr,4")
+    parse.add_argument('--stream_name_0', help = 'stream_name_0',
                        default = "riscv_load_store_rand_instr_stream")
-    parse.add_argument('--stream_freq_opts', help = 'stream_freq_0', default = 4)
+    parse.add_argument('--stream_name_6', help = 'stream_name_6',
+                       default = "riscv_int_numeric_corner_stream")
+    parse.add_argument('--stream_freq_0', help = 'stream_freq_0', default = 4)
+    parse.add_argument('--stream_freq_6', help = 'stream_freq_6', default = 4)
```
udinator, line 423: I think we should dynamically parse these options (if they are passed in) rather than hardcoding each instruction stream to `--directed_instr_[0-9]`, e.g. if a user passes in `--directed_instr_0 riscv_jal_instr,5`, the config object should internally detect that a directed stream is being requested and parse this option and get the correct stream.

pygen/pygen_src/riscv_directed_instr_lib.py
```diff
@@ -11,34 +11,99 @@
...
         self.instr_list[-1].comment = "End %0s" % (self.name)
-        if self.label != "":
-            self.instr_list[0].label = self.label
+        if riscv_directed_instr_stream.label != "":
+            self.instr_list[0].label = riscv_directed_instr_stream.label
             self.instr_list[0].has_label = 1
 
 
+@vsc.randobj
+class riscv_jal_instr(riscv_rand_instr_stream):
```
udinator, line 45: just to confirm, when you run tests with this stream enabled, are they able to compile/simulate with ISS successfully?

pygen/pygen_src/riscv_instr_gen_config.py
aneels3: Got it @udinator 
I will fix this ASAP.

pygen/pygen_src/riscv_directed_instr_lib.py
aneels3: Yes, Checked with both the ISS (Spike and OVPSim).

pygen/pygen_src/riscv_asm_program_gen.py
aneels3: Yes, there will be some performance issues as memory usage will be more.
To fix the label issue we have used the deep copy here.

pygen/pygen_src/riscv_directed_instr_lib.py
udinator: great!

pygen/pygen_src/riscv_instr_gen_config.py
```diff
@@ -411,11 +411,13 @@ def parse_args():
     parse.add_argument('--asm_test_suffix', help = 'asm_test_suffix', default = "")
     parse.add_argument('--march_isa', help = 'march_isa', default = [],
                        choices = [i.name for i in riscv_instr_group_t], nargs = '*')
-    parse.add_argument('--directed_instr_0', help = 'directed_instr_0',
-                       default = "riscv_int_numeric_corner_stream,4")
-    parse.add_argument('--stream_name_opts', help = 'stream_name_0',
-                       default = "riscv_load_store_rand_instr_stream")
-    parse.add_argument('--stream_freq_opts', help = 'stream_freq_0', default = 4)
+    for i in range(7):
```
udinator, line 414: Ideally this should be `for i in range(self.max_directed_instr_stream_seq)` to allow for custom amounts of directed sequences (what if the user inputs more than 7 directed streams?).
To do this though, `parse_args` would need to be moved inside of `riscv_instr_gen_config` object - is this possible?
Possibly you could call this `parse_args` function and do all the `args_dict` setup at the end of `riscv_instr_gen_config.__init__()` .

pygen/pygen_src/riscv_asm_program_gen.py
udinator: that makes sense to me, we can always use a workaround if the performance becomes unacceptable.

pygen/pygen_src/riscv_instr_gen_config.py
pvipsyash: Yes, It is possible. Will fix this

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -596,17 +596,17 @@ def get_directed_instr_stream(self):
             arg = "directed_instr_{}".format(i)
             stream_name_opts = "stream_name_{}".format(i)
             stream_freq_opts = "stream_freq_{}".format(i)
-            if(arg in args):
-                val = args_dict[arg]
+            if(arg in cfg.argv):
```
udinator, line 599: I think you can iterate over `cfg.args_dict` directly, iterating over a `Namespace` object will result in errors.

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -596,17 +596,17 @@ def get_directed_instr_stream(self):
...
                 opts = val.split(",")
                 if(len(opts) != 2):
                     logging.critical(
                         "Incorrect directed instruction format : %0s, expect: name,ratio", val)
                 else:
                     self.add_directed_instr_stream(opts[0], int(opts[1]))
-            elif(stream_name_opts in args and stream_freq_opts in args):
-                stream_name = args_dict[stream_name_opts]
-                stream_freq = args_dict[stream_freq_opts]
+            elif(stream_name_opts in cfg.argv and stream_freq_opts in cfg.argv):
```
udinator, line 607: same as above, should be `elif(stream_name_opts in cfg.args_dict and stream_freq_opts in cfg.args_dict):`

pvipsyash: will fix this one

hodjat91 (approved):
Looks good to me!

PR #691, opened 2020-08-28
[pygen/riscv_instr_stream] Fix ebreak generation

udinator:
Fixes a subtle bug in ebreak generation that would prevent it from being
generated in the debug ROM under a specific parameter combination.

Mirrors #689.

Signed-off-by: Udi Jonnalagadda <udij@google.com>

### DIFF
--- a/pygen/pygen_src/riscv_instr_stream.py
+++ b/pygen/pygen_src/riscv_instr_stream.py
@@ -215,9 +215,15 @@ class riscv_rand_instr_stream(riscv_instr_stream):
             exclude_instr.append(riscv_instr_name_t.C_ADDI16SP.name)
             exclude_instr.append(riscv_instr_name_t.C_LWSP.name)
             exclude_instr.append(riscv_instr_name_t.C_LDSP.name)
-        if is_in_debug and (not cfg.enable_ebreak_in_debug_rom):
-            exclude_instr.append(riscv_instr_name_t.EBREAK.name)
-            exclude_instr.append(riscv_instr_name_t.C_EBREAK.name)
+        # Post-process the allowed_instr and exclude_instr lists to handle
+        # adding ebreak instructions into the debug ROM.
+        if is_in_debug:
+            if (cfg.no_ebreak and cfg.enable_ebreak_in_debug_rom):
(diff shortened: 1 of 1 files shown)


pygen/pygen_src/riscv_instr_stream.py
```diff
@@ -215,9 +215,15 @@ def randomize_instr(self, instr, is_in_debug = 0):
...
             exclude_instr.append(riscv_instr_name_t.C_LDSP.name)
-        if is_in_debug and (not cfg.enable_ebreak_in_debug_rom):
-            exclude_instr.append(riscv_instr_name_t.EBREAK.name)
-            exclude_instr.append(riscv_instr_name_t.C_EBREAK.name)
+        # Post-process the allowed_instr and exclude_instr lists to handle
+        # adding ebreak instructions into the debug ROM.
+        if is_in_debug:
+            if (cfg.no_ebreak and cfg.enable_ebreak_in_debug_rom):
+                allowed_instr.append(riscv_instr_name_t.EBREAK.name)
+                allowed_instr.append(riscv_instr_name_t.C_EBREAK.name)
```
hodjat91, line 223: Sounds good to me, just a minor comment: Instead of `appending` once for each item, better to call `extend` and add all items together:
```suggestion
                allowed_instr.extend([riscv_instr_name_t.EBREAK.name,
                                    riscv_instr_name_t.C_EBREAK.name])
```

pygen/pygen_src/riscv_instr_stream.py
```diff
@@ -215,9 +215,15 @@ def randomize_instr(self, instr, is_in_debug = 0):
...
+            elif (not cfg.no_ebreak and not cfg.enable_ebreak_in_debug_rom):
+                exclude_instr.append(riscv_instr_name_t.EBREAK.name)
+                exclude_instr.append(riscv_instr_name_t.C_EBREAK.name)
```
hodjat91, line 226: and same here.

pygen/pygen_src/riscv_instr_stream.py
udinator: good point, will change this

udinator:
Latest force push changes all `append` calls to `extend`, as per the suggestions.

PR #694, opened 2020-09-02
[Pygen]: minor fixes

hodjat91:
This PR fixes a few issues in riscv_asm_program_gen and riscv_instr_gen_config

PR #695, opened 2020-09-02
[Pygen]: Pyflow integration

hodjat91:
This PR integrates pyflow instruction generator into the overall run flow and makes the required modifications on the pyflow instruction generator side.

### DIFF
--- a/pygen/pygen_src/riscv_asm_program_gen.py
+++ b/pygen/pygen_src/riscv_asm_program_gen.py
@@ -574,8 +574,7 @@ class riscv_asm_program_gen:
         pass
 
     def gen_test_file(self, test_name):
-        subprocess.run(["mkdir", "-p", "out/asm_tests"])
-        file = open("./out/asm_tests/{}".format(test_name), "w+")
+        file = open(test_name, "w+")
         for items in self.instr_stream:
             file.write("{}\n".format(items))
 
--- a/pygen/pygen_src/riscv_instr_gen_config.py
+++ b/pygen/pygen_src/riscv_instr_gen_config.py
@@ -416,6 +416,11 @@ def parse_args():
     parse.add_argument('--stream_name_opts', help = 'stream_name_0',
                        default = "riscv_load_store_rand_instr_stream")
     parse.add_argument('--stream_freq_opts', help = 'stream_freq_0', default = 4)
+    parse.add_argument('--start_idx', help='start index', type=int, default=0)
+    parse.add_argument('--asm_file_name', help='asm file name',
+                       default="riscv_asm_test")
+    parse.add_argument('--log_file_name', help='log file name',
+                       default="")
     # TODO
     '''
     if ($value$plusargs("tvec_alignment=%0d", tvec_alignment)) begin
--- a/pygen/pygen_src/test/riscv_instr_base_test.py
+++ b/pygen/pygen_src/test/riscv_instr_base_test.py
@@ -12,8 +12,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 """
 
 import sys
-sys.path.append("../../")
-from pygen_src.riscv_instr_gen_config import cfg  # NOQA
+sys.path.append("pygen/")
+from pygen_src.riscv_instr_gen_config import cfg, args  # NOQA
 from pygen_src.isa.rv32i_instr import *  # NOQA
 from pygen_src.isa.riscv_instr import riscv_instr_ins  # NOQA
 from pygen_src.riscv_asm_program_gen import riscv_asm_program_gen  # NOQA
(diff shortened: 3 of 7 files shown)


pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -116,8 +116,8 @@ def gen_program(self):
             self.gen_program_end(hart)
             for hart in range(cfg.num_of_harts):
                 self.gen_data_page_begin(hart)
-                if(cfg.no_data_page):
-                    self.gen_data_page()
+                if not cfg.no_data_page:
```
udinator, line 119: I think this change is already in #694?

pygen/pygen_src/riscv_instr_gen_config.py
```diff
@@ -111,16 +111,16 @@ def __init__(self, argv):
         self.num_of_harts = argv.num_of_harts
         self.fix_sp = argv.fix_sp
         self.use_push_data_section = argv.use_push_data_section
-        self.boot_mode_opts = argv.boot_mode_opts
+        self.boot_mode_opts = argv.boot_mode
```
udinator, line 114: same here, I think the `boot_mode` and `init_privileged_mode` fixes are in #694?

pygen/pygen_src/test/riscv_instr_base_test.py
```diff
@@ -12,22 +12,31 @@
...
+                self.asm_file_name = "{}.{}".format(self.asm_file_name,
+                                                    cfg.asm_test_suffix)
+            test_name = "{}_{}.S".format(self.asm_file_name,
+                                         _ + self.start_idx)
+            asm.get_directed_instr_stream()
+            asm.gen_program()
+            asm.gen_test_file(test_name)
+
+riscv_instr_base_test = riscv_instr_base_test()
+riscv_instr_base_test.run_phase()
```
udinator, line 42: looks like you need to add newline at the end of file?

pygen/pygen_src/riscv_asm_program_gen.py
hodjat91: Yes, this PR is actually just a superset of the #694, as it's needed for functionality of the code.

pygen/pygen_src/test/riscv_instr_base_test.py
hodjat91: I'll add the new line at the end of the file. However, it's not a big deal in python. Basically what it means is that the last byte in the file is not a newline, and it's for Diff purposes in git:  Diff has to output a newline anyway, or the result would be harder to read or process automatically.

pygen/pygen_src/riscv_asm_program_gen.py
udinator: ah got it, please make sure to rebase once #694 goes in

pygen/pygen_src/riscv_instr_gen_config.py
udinator: edit: resolved above

hodjat91: That's true. We will only merge this current PR when it is approved. Previous PR just meant to be small fixes to the pyflow generator, and I wanted to keep it separate so it would be easier for perfectVIP team to review it.

pygen/pygen_src/riscv_asm_program_gen.py
hodjat91: Sure, will do.

yaml/simulator.yaml
```diff
@@ -135,3 +135,8 @@
   sim:
     cmd: >
       xrun -R -xmlibdirpath <out> <sim_opts> -svseed <seed> -svrnc rand_struct -nokey
+
+- tool: pyflow
+  sim:
+    cmd: >
+      python3
```
udinator, line 142: I'm not quite sure we should leave this command in the YAML file as just `python3` - can we maybe do something like this?
```console
cmd: >
  python3 <cwd>/pygen/pygen_src/test/<test_name>.py <sim_opts> ...
```
where `<cwd>` is the root of the riscv-dv repo?

I think this way we can let the section on line 278 of `run.py` in this PR just deal with assembling the cwd, test_name, any external sim_opts, and the other 4 'default' arguments - while still making it clear to anyone looking at `simulator.yaml` what the general idea of the Python tool command is doing. Wdyt?

hodjat91: I actually also thought of it, having just "python3" as the sim cmd in the `simulator.yaml` doesn't look good, and it also doesn't provide enough information about this simulation method. I agree with you, and will fix it shortly.

run.py
```diff
@@ -220,13 +225,14 @@ def run_csr_test(cmd_list, cwd, csr_file, isa, iterations, lsf_cmd,
     run_cmd(cmd, timeout_s, debug_cmd = debug_cmd)
 
 
-def do_simulate(sim_cmd, test_list, cwd, sim_opts, seed_gen, csr_file,
-                isa, end_signature_addr, lsf_cmd, timeout_s, log_suffix,
+def do_simulate(sim_cmd, simulator, test_list, cwd, sim_opts, seed_gen,
+                csr_file,isa, end_signature_addr, lsf_cmd, timeout_s, log_suffix,
```
udinator, line 229: nit: `csr_file,isa` -> `csr_file, isa`

run.py
```diff
@@ -71,31 +71,36 @@ def get_generator_cmd(simulator, simulator_yaml, cov, exp, debug_cmd):
...
+    simulator_yaml : RTL/pyflow simulator configuration file in YAML format
     cov            : Enable functional coverage
     exp            : Use experimental version
     debug_cmd      : Produce the debug cmd log without running
 
   Returns:
-    compile_cmd    : RTL simulator command to compile the instruction generator
-    sim_cmd        : RTL simulator command to run the instruction generator
+    compile_cmd    : RTL/pyflow simulator command to compile the instruction
+    generator
```
udinator, line 82: can you indent `generator` to match indentation with `RTL/pyflow` ? both here and below

run.py
```diff
@@ -260,28 +266,42 @@ def do_simulate(sim_cmd, test_list, cwd, sim_opts, seed_gen, csr_file,
...
+            test_cnt = iterations - i * batch_size
+          if simulator == "pyflow":
+            sim_cmd = re.sub("<test_name>", test['gen_test'], sim_cmd)
+            cmd = lsf_cmd + " " + sim_cmd.rstrip() + \
+                  (" --num_of_tests={} ".format(test_cnt)) + \
+                  ("--start_idx={} ".format(i * batch_size)) + \
+                  ("--asm_file_name={}/asm_tests/{} ".format(
+                    output_dir, test['test'])) + \
+                  ("--log_file_name={}/sim_{}_{}{}.log ".format(
+                    output_dir, test['test'], i, log_suffix))
```
udinator, line 286: nit: instead of `.format(...)` can you use the `%` syntax for conformity?

taoliug (approved):

udinator (approved):

PR #701, opened 2020-09-09
[Pygen]: Fix minor issues

aneels3:
This PR fixes minor issues that were generated after the merging of #688 with the overall end-to-end flow.


hodjat91 (approved):
Thanks @aneels3 , I was actually about to make these changes as argv now is an attribute of cfg. LGTM!

PR #703, opened 2020-09-10
[Pygen]: Fix logging issue along with other minor fixes

hodjat91:


pygen/pygen_src/isa/riscv_instr.py
```diff
@@ -156,9 +158,9 @@ def create_csr_filter(self, cfg):
         elif cfg.enable_access_invalid_csr_level:
             self.include_reg = cfg.invalid_priv_mode_csrs
         else:
-            if cfg.init_privileged_mode == "MACHINE_MODE":      # Machine Mode
+            if cfg.init_privileged_mode == privileged_mode_t.MACHINE_MODE:      # Machine Mode
                 self.include_reg.append("MSCRATCH")
-            elif cfg.init_privileged_mode == "SUPERVISOR_MODE":  # Supervisor Mode
+            elif cfg.init_privileged_mode == privileged_mode_t.SUPERVISOR_MODE:  # Supervisor Mode
```
hodjat91, line 163: Hi @aneels3 ,
I believe above three lines should use `priviledged_mode_t.X_MODE` instead of string `"X_MODE"`. Would you please confirm it?

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -599,6 +599,9 @@ def get_directed_instr_stream(self):
                 val = cfg.args_dict[arg]
                 opts = val.split(",")
                 if(len(opts) != 2):
+                    # Do we need this logging.critical here for all iterations in the loop?
+                    # Assuming directed_instr_0=riscv_int_numeric_corner_stream,4 in simulation 
+                    # command, this will generate (max-1) critical messages.
```
hodjat91, line 604: @aneels3 : Please read my comment. This will generate `max - 1` critical messages.

udinator: Can you add a `sys.exit` call after the critical logging message? This should be treated as a fatal error case.

hodjat91: Yeah that makes sense, but I guess my point is that a critical logging message (and/or `sys.exit`) shouldn't happen for "every single iteration", right? I think an additional `if` statement is needed (that checks if `opts` is not empty).

hodjat91: Adding `sys.exit` to the current design leads to exit the program even if the simulation command is valid and correct.

udinator: Not quite sure I understand - in the example of `directed_instr_0=riscv_int_numeric_corner_stream,4` passed in from cmdline, `opts` should be set to `["riscv_int_numeric_corner_stream", "4"]`, right?
The `len(opts)` check will ensure that the arguments are formatted correctly (which in this case they should be), so no  critical message will be generated.

hodjat91: Sorry Udi if my point wasn't clear, basically here is what happens: the `if(arg in cfg.args_dict):` is taken for all 20 iterations (`cfg.max_directed_instr_stream_seq` is 20 and according to lines 422-429 of `riscv_instr_gen_config.py`, they're all added as command line arguments): like you said, in the example of `directed_instr_0=riscv_int_numeric_corner_stream,4` passed in from cmdline, in the first iteration of the `for` loop, opts is set to `["riscv_int_numeric_corner_stream", "4"]`. However, `directed_instr_1` to `directed_instr_19` are also members of `cfg.args_dict`. Hence, `if(arg in cfg.args_dict):` is taken in all 20 iterations (first iteration is non-empty and valid, but for the next 19 iterations, `cfg.args_dict[arg]` is empty). Easy solution to this is to change the  `if(arg in cfg.args_dict):` to `if(cfg.args_dict[arg]):`.

udinator: Ahh, I see what you mean - yes that makes sense to me.
I agree with the proposed fix that uses `if cfg.args_dict[arg]`.

pygen/pygen_src/isa/riscv_instr.py
aneels3: Hi @hodjat91 
Thanks for this. I forgot to do changes for this in the previous #701 PR.
LGTM.

udinator (approved):

PR #704, opened 2020-09-10
Workaround of the SV compilation problem caused by assigning the const array variable with the empty concatenation.

dariuszst:
There "initial custom CSR support" commit introduced an empty constant "custom_csr":
https://github.com/google/riscv-dv/blob/0645eeaf3e1cd5927c32372b3ae142323ca3e9e8/target/rv32imc/riscv_core_setting.sv#L117-L123

The initialization of a constant variable with the empty array concatenation expression causes compilation problem on some tools, e.g. Riviera-PRO. 
It seems that this empty const variable is not used in the generator code - is it really necessary? The creation of empty const array is a little strange - what is a reason for creation of such variable?

This PR contains a workaround for Riviera-PRO that is identical as this  already existing for DSIM.

### DIFF
--- a/target/ml/riscv_core_setting.sv
+++ b/target/ml/riscv_core_setting.sv
@@ -115,15 +115,7 @@ const privileged_reg_t implemented_csr[] = {
 };
 
 // Implementation-specific custom CSRs
-`ifdef DSIM
 bit [11:0] custom_csr[] = {
-`else
-`ifdef _VCP //DST713
-bit [11:0] custom_csr[] = {
-`else
-const bit [11:0] custom_csr[] = {
-`endif
-`endif
 };
 
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


taoliug:
I think it's ok to remove the "const", @udinator, any thought?

udinator:
Agree, I think simply removing the const identifier will be ok here.

dariuszst:
> Agree, I think simply removing the const identifier will be ok here.

I'think this would be the simplest and the clearest solution.

taoliug:
@dariuszst , can you update the PR to remove the "const" ? thanks

dariuszst:
> @dariuszst , can you update the PR to remove the "const" ? thanks

OK. I've removed const and WA for DSIM and Riviera-PRO.

taoliug (approved):

PR #708, opened 2020-09-11
[PyFlow]: FC integration

hodjat91:
This PR integrates functional coverage side of pyflow into cov.py and makes required modifications.

### DIFF
--- a/cov.py
+++ b/cov.py
@@ -76,31 +76,39 @@ def sim_cov(out, cfg, cwd, opts_vec, opts_cov, csv_list):
   test_name = "riscv_instr_cov_test"
   base_sim_cmd = ("python3 %s/run.py --simulator %s --simulator_yaml %s --noclean "
                   "--so -o %s --cov -tl %s %s "
-                  "-tn %s --steps gen --sim_opts \"<trace_csv_opts> %s %s\" " %
+                  "-tn %s --steps gen --sim_opts \"<trace_csv_opts> %s %s <visualization>\" " %
                   (cwd, argv.simulator, argv.simulator_yaml, out, argv.testlist,
                    argv.opts, test_name, opts_vec, opts_cov))
+  if argv.simulator == "pyflow" and argv.enable_visualization: 
+    base_sim_cmd = re.sub("<visualization>", "--enable_visualization", base_sim_cmd)
+  else:
+    base_sim_cmd = re.sub("<visualization>", "", base_sim_cmd)
   if argv.target:
     base_sim_cmd += (" --target %s" % argv.target)
   if argv.custom_target:
     base_sim_cmd += (" --custom_target %s" % argv.custom_target)
   if argv.stop_on_first_error:
     base_sim_cmd += (" --stop_on_first_error")
-  file_idx = 0
-  trace_idx = 0
   trace_csv_opts = ""
   batch_cnt = 1
   sim_cmd_list = []
   if argv.batch_size > 0:
-    batch_cnt = (len(csv_list) + argv.batch_size - 1)/ argv.batch_size;
+    batch_cnt = (len(csv_list) + argv.batch_size - 1) / argv.batch_size
     logging.info("Batch size: %0d, Batch cnt:%0d" % (argv.batch_size, batch_cnt))
   for i in range(len(csv_list)):
     file_idx = 0
     trace_idx = i
     if argv.batch_size > 0:
-      file_idx = i / argv.batch_size;
-      trace_idx = i % argv.batch_size;
-    trace_csv_opts += (" +trace_csv_%0d=%s" % (trace_idx, csv_list[i]))
-    if ((i == len(csv_list)-1) or ((argv.batch_size > 0) and (trace_idx == argv.batch_size-1))):
+      file_idx = i / argv.batch_size
+      trace_idx = i % argv.batch_size
+    if argv.simulator == "pyflow":
+      if i == 0:
+        trace_csv_opts += (" --trace_csv=%s" % csv_list[i])
+      else:
+        trace_csv_opts += (",%s" % csv_list[i])
+    else:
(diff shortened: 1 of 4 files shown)


cov.py
```diff
@@ -76,31 +76,39 @@ def sim_cov(out, cfg, cwd, opts_vec, opts_cov, csv_list):
   test_name = "riscv_instr_cov_test"
   base_sim_cmd = ("python3 %s/run.py --simulator %s --simulator_yaml %s --noclean "
                   "--so -o %s --cov -tl %s %s "
-                  "-tn %s --steps gen --sim_opts \"<trace_csv_opts> %s %s\" " %
+                  "-tn %s --steps gen --sim_opts \"<trace_csv_opts> %s %s <visualization>\" " %
                   (cwd, argv.simulator, argv.simulator_yaml, out, argv.testlist,
                    argv.opts, test_name, opts_vec, opts_cov))
+  if argv.simulator == "pyflow" and argv.enable_visualization:
```
udinator, line 82: could something like this work instead?
```python
if argv.simulator == "pyflow" and argv.enable_visualization:
  base_sim_cmd += (" --enable_visualization")
```
it might simplify the logic a little bit, wdyt?

cov.py
```diff
@@ -76,31 +76,39 @@ def sim_cov(out, cfg, cwd, opts_vec, opts_cov, csv_list):
...
   if argv.stop_on_first_error:
     base_sim_cmd += (" --stop_on_first_error")
-  file_idx = 0
-  trace_idx = 0
   trace_csv_opts = ""
   batch_cnt = 1
   sim_cmd_list = []
   if argv.batch_size > 0:
-    batch_cnt = (len(csv_list) + argv.batch_size - 1)/ argv.batch_size;
+    batch_cnt = (len(csv_list) + argv.batch_size - 1)/ argv.batch_size
```
udinator, line 96: small nit: needs a space before the `/` sign

pygen/pygen_src/test/riscv_instr_cov_test.py
```diff
@@ -15,34 +15,39 @@
...
+        self.argv = self.parse_args()
+        logging.basicConfig(filename='{}'.format(self.argv.log_file_name),
+                            filemode='w',
+                            format="%(filename)s %(lineno)s %(levelname)s %(message)s",
+                            level=logging.DEBUG)
 
     def run_phase(self):
+        trace_files = self.argv.trace_csv.split(",")
+        for item in trace_files:
+            self.csv_trace.append(item.split("_", 1)[1])
```
udinator, line 45: to clarify, this is needed to remove the `_[i]` suffix, correct?

hodjat91: Yes, it remove the `[i]_` prefix (although it's not needed at all, but I wanted to follow the FC on SV side).

cov.py
hodjat91: Actually `--enable_visualization` is not known by the `run.py` (it's an command line argument of `riscv_instr_cov_test`). Hence, it has to come right after `<trace_csv_opts> %s %s` and **inside double quotes**, that's why I've used `re.sub("<visualization>", "--enable_visualization", base_sim_cmd)`. `base_sim_cmd += (" --enable_visualization")` won't work because it just appends `--enable_visualization` to the end of command.

udinator: yup you're right, I missed the double quotes the first time through. thanks for pointing that out!

cov.py
```diff
@@ -76,31 +76,39 @@ def sim_cov(out, cfg, cwd, opts_vec, opts_cov, csv_list):
...
-      trace_idx = i % argv.batch_size;
-    trace_csv_opts += (" +trace_csv_%0d=%s" % (trace_idx, csv_list[i]))
-    if ((i == len(csv_list)-1) or ((argv.batch_size > 0) and (trace_idx == argv.batch_size-1))):
+      file_idx = i / argv.batch_size
+      trace_idx = i % argv.batch_size
+    if argv.simulator == "pyflow":
+      if i == 0:
+        trace_csv_opts += (" --trace_csv=%0d_%s" % (trace_idx, csv_list[i]))
+      else:
+        trace_csv_opts += (",%0d_%s" % (trace_idx, csv_list[i]))
```
udinator, line 108: I'm not sure if it's needed here to prefix the csv file path with the index since the CSV files will already contain an index number in their pathname e.g. `riscv_jump_stress_test.8.csv` so they should be unique already - it seems like the cov_test.py strips these prefixes immediately anyways.

pygen/pygen_src/test/riscv_instr_cov_test.py
udinator: just left a comment above about this, but I think the indexing can be removed altogether, the SV side only uses the index to uniquely name plusargs which the cov_test.sv can pick up during runtime.

cov.py
hodjat91: Yes, that is true. However, I think keeping these prefix indices is a good idea in a case where we have different tests: say we have `riscv_jump_stress_test.0.csv` and `riscv_arithmeric_basic_test.1.csv`. In the case of multiple type test files, this index can act as a global index for all tests. What do you think?

hodjat91: I agree that in the current scenario it's redundant though.

pygen/pygen_src/test/riscv_instr_cov_test.py
hodjat91: Ah got it, that makes sense. I'll remove it then.

cov.py
hodjat91: You're right :-) It's not needed as we already have all the CSV files printed out. I've removed the indices part and pushed a new commit reflecting the changes.

udinator (approved):
lgtm!

taoliug (approved):

Issue #709, opened 2020-09-14
Compliation error for test case riscv_rand_instr_test with instr_cnt of 200000


suhasskrishanmurthy:
Hi, 
  I changed instr_cnt to 200000 in file ./yaml/base_testlist.yaml. Now ASM files are being genegerated but compilation is throwing error(linker error).
Please kindly tell me, what is that I am doing wrong.
Command to Run is,
 **$python3 run.py --custom_target ./target/rv64im --isa rv64im --mabi lp64 --test riscv_rand_instr_test                --simulator questa**
The report generated by riscv-dv is below for riscv_rand_instr_test test case.

Mon, 14 Sep 2020 10:19:24 INFO     Creating output directory: out_2020-09-14
Mon, 14 Sep 2020 10:19:24 INFO     Processing regression test list : ./target/rv64im/testlist.yaml, test: riscv_rand_instr_test
Mon, 14 Sep 2020 10:19:24 INFO     Processing regression test list : /home/ariane/tandem/test_vectors/riscv-dv/target/rv32imc/testlist.yaml, test: riscv_rand_instr_test
Mon, 14 Sep 2020 10:19:24 INFO     Processing regression test list : /home/ariane/tandem/test_vectors/riscv-dv/yaml/base_testlist.yaml, test: riscv_rand_instr_test
Mon, 14 Sep 2020 10:19:24 INFO     Found matched tests: riscv_rand_instr_test, iterations:2
Mon, 14 Sep 2020 10:19:24 INFO     Processing simulator setup file : /home/ariane/tandem/test_vectors/riscv-dv/yaml/simulator.yaml
Mon, 14 Sep 2020 10:19:24 INFO     Found matching simulator: questa
Mon, 14 Sep 2020 10:19:24 INFO     Building RISC-V instruction generator
Mon, 14 Sep 2020 10:19:40 INFO     Running RISC-V instruction generator
Mon, 14 Sep 2020 10:19:40 INFO     Generating 2 riscv_rand_instr_test
Mon, 14 Sep 2020 10:19:40 INFO     Running riscv_rand_instr_test with 1 batches
Mon, 14 Sep 2020 10:19:40 INFO     Running riscv_rand_instr_test, batch 1/1, test_cnt:2
Mon, 14 Sep 2020 10:27:26 INFO     Compiling out_2020-09-14/asm_tests/riscv_rand_instr_test_0.S

/tmp/ccKK42Yk.o: In function `mepc_setup':
[...]

PR #710, opened 2020-09-15
fix jumps to `test_done` and `init_[m/s/u]_mode`

udinator:
From #709, it appears there is an issue jumping to `init_machine_mode`
when generating a large (200k instructions) program.
This is because we currently use a standard `j` instruction to do the
jump which only has a 20-bit address range, so this range is exceeded in
a program of this size.
Similar issues arise when jumping to `test_done` and `write_tohost` from 
various exception handlers.

The solution is as follows:

1) Move the `init_[m/s/u]_mode` section to the top of the program to
prevent us from having to jump there in the first place - this does not
cause any side effects.
2) Change the jump mechanisms to `test_done` and `write_tohost` from `j test_done` to
preloading the address of `test_done` into a register and using a `jalr`
instruction to jump to this address - this allows us to cover a full
XLEN-bit address space when jumping around the program.

This PR also refactors `gen_ecall_handler()`, and includes the above changes.

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -117,7 +117,10 @@ class riscv_asm_program_gen extends uvm_object;
       instr_stream = {instr_stream, main_program[hart].instr_string_list};
       // If PMP is supported, need to jump from end of main program to test_done section at the end
       // of main_program, as the test_done will have moved to the beginning of the program
-      instr_stream = {instr_stream, $sformatf("%sj test_done", indent)};
+      instr_stream = {instr_stream,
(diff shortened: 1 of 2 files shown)


hodjat91 (approved):
The proposed approach looks good to me!

udinator:
The latest force push makes use of the new jumping mechanism to jump to the `write_tohost` label, as well as refactors the `gen_ecall_handler()` function with related changes.

hodjat91:
Proposed changes make sense to me, LGTM!

taoliug (approved):

PR #711, opened 2020-09-15
Convert python codes to be PEP8 compliant

hodjat91:
This PR converts the python codes in riscv-dv repo to be PEP8 compliant (PR also contains other minor fixes).

### DIFF
--- a/scripts/gen_csr_test.py
+++ b/scripts/gen_csr_test.py
@@ -21,7 +21,6 @@ in the CSR at this point, allowing for the test to self-check in order to
 determine success or failure.
 """
 
-
 """
 To install the bitstring library:
   1) sudo apt-get install python3-bitstring OR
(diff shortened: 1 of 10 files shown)


scripts/gen_csr_test.py
```diff
@@ -35,325 +34,362 @@
...
+    rv_string = "rv{}".format(str(xlen))
+    csrs = {}
+    with open(csr_file, "r") as c:
+        csr_description = yaml.safe_load(c)
+        for csr_dict in csr_description:
+            csr_name = csr_dict.get("csr")
+            csr_address = csr_dict.get("address")
+            assert (
+                        rv_string in csr_dict), "The {} CSR must be configured for rv{}".format(
+                csr_name, str(rv))
```
udinator, line 73: is it possible to fix this formatting? I think it looks a bit strange like this

hodjat91: Good catch, I fixed this.

udinator (approved):
LGTM

PR #732, opened 2020-10-19
[Pygen]: Add support for RV32F and RV32D

aneels3:
This PR adds support for Floating-point instructions.


ShraddhaDevaiya:
@googlebot I consent.

pvipsyash:
@googlebot I consent.

aneels3:
Hi @udinator and @hodjat91,
General Updates:
- Rename the target from rv32imfc to rv32imfdc for support of both F and D extension.
- Add the stable version of PyVSC to requirement.txt as I am experiencing some [issue](https://github.com/fvutils/pyvsc/issues/43) with the new release. This will be reverted in a future commit.

Regards,
Anil

ishita71:
@googlebot I consent.

pygen/pygen_src/isa/riscv_floating_point_instr.py
```diff
@@ -0,0 +1,150 @@
...
+        self.fs3 = vsc.rand_enum_t(riscv_fpr_t)
+        self.fd = vsc.rand_enum_t(riscv_fpr_t)
+        self.has_fs1 = vsc.bit_t(1)
+        self.has_fs2 = vsc.bit_t(1)
+        self.has_fs3 = vsc.bit_t(1)
+        self.has_fd = vsc.bit_t(1)
+        self.has_fs1 = 1
+        self.has_fs2 = 1
+        self.has_fs3 = 0
+        self.has_fd = 1
```
udinator, line 36: looks like these are duplicated?

pygen/pygen_src/isa/riscv_floating_point_instr.py
```diff
@@ -0,0 +1,150 @@
...
+            elif self.instr_name.name in ['FMV_X_W', 'FMV_X_D', 'FCVT_W_S', 'FCVT_WU_S',
+                                          'FCVT_L_S', 'FCVT_LU_S', 'FCVT_L_D', 'FCVT_LU_D',
+                                          'FCVT_LU_S', 'FCVT_W_D', 'FCVT_WU_D']:
+                self.has_fd = 0
+                self.has_rd = 1
+            elif self.instr_name.name in ['FMV_W_X', 'FMV_D_X', 'FCVT_S_W', 'FCVT_S_WU',
+                                          'FCVT_S_L', 'FCVT_D_L', 'FCVT_S_LU', 'FCVT_D_W',
+                                          'FCVT_D_LU', 'FCVT_D_WU']:
+                self.has_rs1 = 1
+                self.fs1 = 0
```
udinator, line 96: nit: this should be `self.has_fs1`

pygen/pygen_src/riscv_reg.py
```diff
@@ -0,0 +1,131 @@
...
+    def zero_reserved_field_c(self):
+        with vsc.implies(self.access_type == reg_field_access_t.WPRI):
+            self.val == 0
+
+    @vsc.constraint
+    def hardwired_fld_c(self):
+        with vsc.implies(self.hard_wired == 1):
+            self.val == self.reset_val
+
+    def covert2string(self):
```
udinator, line 46: nit: `covert2string` -> `convert2string`

pygen/pygen_src/riscv_reg.py
```diff
@@ -0,0 +1,131 @@
...
+        return pkg_ins.format_string("{} bit_width:{} val:{} type:{}".format(
+            self, self.bit_width, hex(self.val), self.access_type))
+
+    def post_randomize(self):
+        mask = vsc.bit_t(rcs.XLEN)
+        mask = 1
```
udinator, line 52: this value should actually be `XLEN'1`, instead of 1

pygen/pygen_src/riscv_utils.py
```diff
@@ -38,18 +35,64 @@ def factory(obj_of):
         sys.exit(1)
 
 
-def get_object(instr):
+def get_object(instr_name):
```
udinator, line 38: would it be possible to to somewhat simplify this function? I think manually checking equality of all the targets and for each individual extension doesn't scale well to when we start adding the 64-bit extensions and so on.

an initial idea I thought of (may not be feasible) would be to first reverse-index the `riscv_instr::instr_group` list to find which instruction group `instr_name` belongs to (e.g. `RV32I`, `RV32C`, so on), then that group name could used used to create the name of the correct instruction package to import.
I think that would save a lot of the effort currently required to manually import every possible package and try creating the instruction object. WDYT?

target/rv32imfdc/riscv_core_setting.sv
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http: // www.apache.org / licenses / LICENSE - 2.0
```
udinator, line 8: nit: can you please remove the unneeded spaces in the URL?

target/rv32imfdc/testlist.yaml
```diff
@@ -0,0 +1,81 @@
...
+# sim_opts        : Simulation options passed to the instruction generator
+# no_post_compare : Enable/disable comparison of trace log and ISS log (Optional)
+# compare_opts    : Options for the RTL & ISS trace comparison
+# gcc_opts        : gcc compile options
+# --------------------------------------------------------------------------------
+
+- import: <riscv_dv_root>/target/rv32imc/testlist.yaml
+
+
+- test: riscv_floating_point_arithmetic_test
```
udinator, line 36: just to confirm, if you run a `rv32imfdc` regression with Spike and OVPsim cross comparison, are there any failures?

requirements.txt
```diff
@@ -7,6 +7,6 @@ sphinx-issues
 sphinx_rtd_theme
 rst2pdf
 flake8
-pyvsc
+pyvsc==0.2.3.20201006.1
```
hodjat91, line 10: Is there any issue with the latest version (0.2.5)?

pygen/pygen_src/riscv_utils.py
hodjat91: I agree with your proposal Udi, and I think it saves lots of effort (and also that would look better in the code rather than manually checking the equality).

pygen/pygen_src/riscv_reg.py
hodjat91: I think this will still show it as XLEN'00..001, but I suggest combine the two lines: mask = vcs.bit_t(rcs.XLEN, 1). WDYT?

pygen/pygen_src/riscv_reg.py
```diff
@@ -0,0 +1,131 @@
...
+                logging.info(self.fld[i].convert2string())
+            logging.critical("Total field length {} != XLEN {}".format(total_len, rcs.XLEN))
+            sys.exit(1)
+        self.val = 0
+        for i in range((len(self.fld) - 1), -1, -1):
+            self.val = (self.val << self.fld[i].bit_width) | self.fld[i].val
+        return self.val
+
+    def add_field(self, fld_name, bit_width, access_type, reset_val = 0):
+        new_fld = fld_name
```
hodjat91, line 86: I think this line is redundant and should be removed (as new_fld will be updated in the next line, and the fld_name will be assigned to the name field later on anyways).

pygen/pygen_src/riscv_reg.py
udinator: whoops sorry, I meant `'1` (in systemverilog land), which would translate to `0xffff_..._ffff

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -338,7 +338,55 @@ def init_gpr(self):
...
+        # shift to upper 32bits
+        for _ in range(2):
+            string = "{}slli x{}, x{}, 16".format(pkg_ins.indent, int_gpr1, int_gpr1)
+            self.instr_stream.append(string)
+        string = "{}li x{}, {}".format(pkg_ins.indent, int_gpr2, imm[31:0])
+        self.instr_stream.append(string)
+        string = "{}or x{}, x{}, x{}".format(pkg_ins.indent, int_gpr2, int_gpr2, int_gpr1)
+        self.instr_stream.append(string)
+        string = "{}fmv.d.x f{}, x{}".format(pkg_ins.indent, int_floating_gpr, int_gpr2)
+        self.instr_stream.append(string)
```
hodjat91, line 377: I think it would look better to give the "string" a more meaningful name (e.g., `li_inst`, `or_inst`, etc) and use `extend` instead of `append` to add the three lines altogether at the same time instead of keep repeating the calls to `append` (`self.instr_stream.extend([li_inst, or_instr, ...])`)

pygen/pygen_src/riscv_reg.py
hodjat91: Ahh got it.

pygen/pygen_src/isa/riscv_floating_point_instr.py
aneels3: Hi @udinator 
Actually, these lines are not duplicated. I have used it here to assign the default value. Since `vsc.bit_t(1)` has default value of 0.
Anyways, I have removed these statements and define these fields as `self.has_fs1 = vsc.bit_t(1, 1)` in order to retain the default values.

target/rv32imfdc/testlist.yaml
aneels3: No, Regression is running fine for both the ISS.

requirements.txt
aneels3: As of now, I have not encountered any issues with the new release.
Reverted this back to the original state.

pygen/pygen_src/riscv_reg.py
aneels3: Correct!
Thanks, @hodjat91

pygen/pygen_src/isa/riscv_floating_point_instr.py
udinator: ahh, that makes sense.
I think it might be more efficient to set the default value at the time of declaration, like you have updated this to.

pygen/pygen_src/riscv_utils.py
aneels3: Hi @udinator and @hodjat91,
I have thought about this before as well but couldn't able to make it generic with the instr_group. As **`cls.instr_group`** list is updated after the **`create_instr`** function call. So, In order to create an instance for the first instruction, I will get an empty **`cls.instr_group`** list and maybe the corresponding group of the previous instr_name for the successive iteration.

I just figure out another approach, possibly a WA to this particular issue.
During the registration of each instruction, I am thinking to pass two parameters instead of one i.e 
```python3
def register(cls, instr_name, instr_group):
    cls.instr_registry[instr_name] = instr_group
```
In order to register each instruction with the corresponding group as a value like above.

So that I can iterate over the **key** and **value** in the **create_instr_list()** function
```python3
def create_instr_list(cls, cfg):
    for isntr_name, instr_group in cls.instr_registry.items():
        instr_inst = cls.create_instr(instr_name, instr_group)
```

and implement the rest of the thing in the create_instr() function.
This WA will not require manual modification when we add new extensions in the coming days. Like I am doing in **riscv_utils.py** currently.
Let me know if it works or not!

hodjat91: Hi @aneels3 

I think your proposed approach should work and is obviously better than fetch things manually (I'm thinking about other alternatives and will let you know if I can come up with a better solution).

udinator (approved):
LGTM

hodjat91 (approved):
LGTM as well! Thanks for the fix, looks much better now.

PR #736, opened 2020-10-23
coverage flow updates

udinator:
This PR does 2 main things to the functional coverage flow
1) Allows the .csv conversion scripts to recognize bit-manipulation pseudo-instructions - currently none are recognized so coverage failures are fairly common.
2) Changes how the 'operand' field of the output CSV is formatted.
Currently for any instructions that have an address (e.g. `rd, rs1(imm)`), the operands are structured as `rd,imm,rs1` in the output CSV file. This doesn't align with how the spec structures operands for these sorts of instructions, so this patch rearranges the operands such that they are displayed as `rd,rs1,imm` in the output CSV.
This has the side effect of standardizing how the CSVs are parsed among Spike, OVPsim, and any cores using riscv-dv (like Ibex).

### DIFF
--- a/src/isa/riscv_instr_cov.svh
+++ b/src/isa/riscv_instr_cov.svh
@@ -268,10 +268,10 @@
       I_FORMAT: begin
         `DV_CHECK_FATAL(operands.size() == 3, instr_name)
         if(category == LOAD) begin
-          // load rd, imm(rs1)
-          rs1 = get_gpr(operands[2]);
-          rs1_value = get_gpr_state(operands[2]);
-          get_val(operands[1], imm);
+          // load rd, imm(rs1) -> rd,rs1,imm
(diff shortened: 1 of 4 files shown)


src/isa/riscv_instr_cov.svh
```diff
@@ -268,10 +268,10 @@
       I_FORMAT: begin
```
taoliug, line 268: Please run a coverage regression to validate this change

udinator: Yep I ran a coverage regression with rv32imcb, and everything is functioning as intended.
rv64.... regressions are not as stable yet (mostly because of V-extension and unsupported B pseudoinstructions), but I have TODO items in this PR that I will address soon.

taoliug (approved):

PR #739, opened 2020-10-30
Pass -fno-extended-identifiers to VCS

rswarbrick:
For some bizarre reason, VCS includes DPI code that has smart
quotes (yes, smart quotes) around some #error macros. Recent GCC
versions (g++ >= 10.2) have got proper support for UTF-8, which is
nice but makes them choke on these input files.

Passing -fno-extended-identifiers tells GCC to pretend that everything
is ASCII, and all is well again!

### DIFF
--- a/yaml/simulator.yaml
+++ b/yaml/simulator.yaml
@@ -21,6 +21,7 @@
              -f <cwd>/files.f -full64
              -l <out>/compile.log
              -LDFLAGS '-Wl,--no-as-needed'
+             -CFLAGS '--std=c99 -fno-extended-identifiers'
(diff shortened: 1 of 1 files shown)


udinator:
Whoops...seems like I duplicated this PR without seeing yours first!
There is a fairly urgent need to get this patch merged and then vendored into Ibex, so I will merge my version of the PR (Weicai has requested that I add some explanation of the flag in a comment for context, seeing that there is a pretty esoteric reason for this patch in the first place) and will credit you as the primary author as you did open a PR much earlier.
Sorry for the confusion!

PR #740, opened 2020-10-30
add VCS compile option for unicode

udinator:
VCS includes smart quotes in some DPI code, making G++ blow up during
compilation.
As a result we need to pass in -fno-extended-identifiers to tell G++ to
pretend that everything is ASCII and not error out during compilation.

Primary author: @rswarbrick
Co-author: @udinator

### DIFF
--- a/yaml/simulator.yaml
+++ b/yaml/simulator.yaml
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# The -CFLAGS option is required as some VCS DPI code contains smart quotes
+# around some preprocessor macros, making G++ throw errors during compilation.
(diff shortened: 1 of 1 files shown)


yaml/simulator.yaml
```diff
@@ -21,6 +21,7 @@
              -f <cwd>/files.f -full64
              -l <out>/compile.log
              -LDFLAGS '-Wl,--no-as-needed'
+             -CFLAGS '--std=c99 -fno-extended-identifiers'
```
weicaiyang, line 24: nit: perhaps good to add some comment for this option

weicaiyang (approved):

yaml/simulator.yaml
udinator: done.

PR #745, opened 2020-11-07
[Pygen]: Add support for RV32B instructions

aneels3:
This PR adds the following contents:
- Bit Manipulation Extension instruction.
- Most of the previous TODO's.
- Fixed inline randomization of randomize_gpr() which was failing with the PyVSC 0.2.6 release. 
**Regressions are checked and verified with SPIKE and OVPsim.**

There is an issue with the size method while using it inside the constraint construct.
This particular issue is blocking the working of Atomic instructions. I will create a separate PR for A extension once this is fixed.
In fact with a workaround, I am able to run the regression for RV32A instructions.


ShraddhaDevaiya:
@googlebot I consent.

ishita71:
@googlebot I consent.

pvipsyash:
@googlebot I consent.

pygen/pygen_src/isa/riscv_b_instr.py
```diff
@@ -0,0 +1,178 @@
...
+    def pre_randomize(self):
+        super().pre_randomize()
+        with vsc.raw_mode():
+            self.rs3.rand_mode = bool(self.has_rs3)
+
+    def set_imm_len(self):
+        if self.format == riscv_instr_format_t.I_FORMAT:
+            if self.category in [riscv_instr_category_t.SHIFT, riscv_instr_category_t.LOGICAL]:
+                if (self.group.name == riscv_instr_group_t.RV64B and
+                        not(self.instr_name in [riscv_instr_name_t.SLLIU_W])):
```
hodjat91, line 65: since the list has only one member, I think it's better for this line to be replaced with:
`self.instr_name != riscv_instr_name_t.SLLIU_W`

pygen/pygen_src/riscv_instr_gen_config.py
```diff
@@ -179,31 +175,32 @@ def __init__(self):
...
-        self.gpr0.not_inside(vsc.rangelist(self.sp, self.tp, self.scratch_reg, self.pmp_reg,
-                                           riscv_reg_t.ZERO, riscv_reg_t.RA, riscv_reg_t.GP))
-        self.gpr1.not_inside(vsc.rangelist(self.sp, self.tp, self.scratch_reg, self.pmp_reg,
-                                           riscv_reg_t.ZERO, riscv_reg_t.RA, riscv_reg_t.GP))
-        self.gpr2.not_inside(vsc.rangelist(self.sp, self.tp, self.scratch_reg, self.pmp_reg,
-                                           riscv_reg_t.ZERO, riscv_reg_t.RA, riscv_reg_t.GP))
-        self.gpr3.not_inside(vsc.rangelist(self.sp, self.tp, self.scratch_reg, self.pmp_reg,
-                                           riscv_reg_t.ZERO, riscv_reg_t.RA, riscv_reg_t.GP))
-        vsc.unique(self.gpr0, self.gpr1, self.gpr2, self.gpr3)
+        with vsc.foreach(self.gpr, idx=True) as i:
```
hodjat91, line 200: nit: space needed: `idx = True`

pygen/pygen_src/riscv_instr_gen_config.py
```diff
@@ -329,36 +322,26 @@ def get_invalid_priv_lvl_csr(self):
...
-                invalid_lvl.append('M')
-                logging.info("usr_mode---")
-                logging.debug(invalid_lvl)
-            else:
-                logging.critical("Unsupported initialization privilege mode")
+        if self.init_privileged_mode == privileged_mode_t.MACHINE_MODE:
+            pass
+        elif self.init_privileged_mode == privileged_mode_t.SUPERVISOR_MODE:
+            invalid_lvl.append('M')
+            logging.info("supr_mode---")
```
hodjat91, line 329: nit: better to be consistent in using `''` vs. `""`: `invalid_lvl.append("M")`. Here as well as few other places below.

pygen/pygen_src/riscv_instr_sequence.py
```diff
@@ -54,9 +54,22 @@ def gen_instr(self, is_main_program, no_branch = 1):
             self.gen_stack_enter_instr()
             self.gen_stack_exit_instr()
 
-    # TODO
     def gen_stack_enter_instr(self):
-        pass
+        allow_branch = 0 if ((self.illegal_instr_pct > 0) or (self.hint_instr_pct > 0)) else 1
```
hodjat91, line 58: Nit: redundant parentheses:  `allow_branch = 0 if (self.illegal_instr_pct > 0 or self.hint_instr_pct > 0) else 1`

pygen/pygen_src/riscv_instr_sequence.py
```diff
@@ -54,9 +54,22 @@ def gen_instr(self, is_main_program, no_branch = 1):
...
+                                                                cfg.max_stack_len_per_program))
+                self.program_stack_len % (rcs.XLEN // 8) == 0
+        except Exception:
+            logging.critical("Cannot randomize program_stack_len")
+            sys.exit(1)
+        self.instr_stack_enter.push_start_label = self.label_name + "_stack_p"
+        self.instr_stack_enter.gen_push_stack_instr(self.program_stack_len,
+                                                    allow_branch = allow_branch)
+        self.instr_stream.instr_list.extend((self.instr_stack_enter.instr_list,
+            self.instr_stream.instr_list))
```
hodjat91, line 72: These two lines won't work as `extend` intends, it will append two separate "lists" to the `self.instr_stream.instr_list`. Just to confirm, is that intentional?  If not, perhaps we should replace it with:
`self.instr_stream.instr_list.extend(self.instr_stack_enter.instr_list)
self.instr_stream.instr_list.extend(self.instr_stream.instr_list)`

pygen/pygen_src/riscv_instr_stream.py
```diff
@@ -181,9 +181,19 @@ def setup_allowed_instr(self, no_branch = 0, no_load_store = 1):
                 riscv_instr.instr_category[riscv_instr_category_t.STORE.name])
         self.setup_instruction_dist(no_branch, no_load_store)
 
-    # TODO
     def randomize_avail_regs(self):
-        pass
+        if(self.avail_regs.size > 0):
```
hodjat91, line 185: Nit: redundant parentheses

pygen/pygen_src/riscv_privileged_common_seq.py
```diff
@@ -102,20 +105,82 @@ def setup_mmode_reg(self, mode, regs):
...
+            self.sie = riscv_privil_reg()
+            self.sie.init_reg(privileged_reg_t.SIE)
+            if cfg.randomize_csr:
+                self.sie.set_val(cfg.sie)
+            self.sie.set_field("UEIE", cfg.enable_interrupt)
+            self.sie.set_field("SEIE", cfg.enable_interrupt)
+            self.sie.set_field("USIE", cfg.enable_interrupt)
+            self.sie.set_field("SSIE", cfg.enable_interrupt)
+            self.sie.set_field("STIE", cfg.enable_interrupt & cfg.enable_timer_irq)
+            self.sie.set_field("UTIE", cfg.enable_interrupt & cfg.enable_timer_irq)
```
hodjat91, line 137: For better readability, perhaps better to create a separate method to do all the `set_fields` (for `self.sstatus` as well as `self.sie`)

pygen/pygen_src/riscv_privileged_common_seq.py
```diff
@@ -102,20 +105,82 @@ def setup_mmode_reg(self, mode, regs):
...
+        fld_name = satp.get_field_by_name("PPN")
+        satp_ppn_mask = hex((2**rcs.XLEN) - 1) >> (rcs.XLEN - fld_name.bit_width)
+        # Load the root page table physical address
+        instrs.append("la x{}, page_table_0".format(cfg.gpr[0]))
+        # Right shift to get PPN at 4k granularity
+        instrs.append("srli x{}, x{}, 12".format(cfg.gpr[0], cfg.gpr[0]))
+        instrs.append("li   x{}, {}".format(cfg.gpr[1], hex(satp_ppn_mask)))
+        instrs.append("and x{}, x{}, x{}".format(cfg.gpr[0], cfg.gpr[0], cfg.gpr[1]))
+        # Set the PPN field for SATP
+        instrs.append("csrs {}, x{} // satp".format(hex(privileged_reg_t.SATP), cfg.gpr[0]))
```
hodjat91, line 186: Could you please use a single `extend` instead of repetitive `appends`?

pygen/pygen_src/riscv_instr_sequence.py
aneels3: Hi @hodjat91 
This is not intensional, In fact, the second list is inappropriate to use here.
Just realize after looking at your comment,
Thanks for this
I will fix it right away,

pygen/pygen_src/riscv_privileged_common_seq.py
aneels3: Hi @hodjat91 
These are itself in a method which is being called from the `enter_privileged_mode` and using `status` and `sie` objects I am calling the `set_fields` method.
Even If I create new methods for these, The line of code and other things would be the same.
What do you think?

aneels3: Hi @hodjat91 
I have changed this as you suggested.
Please do have a look!

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -95,7 +94,8 @@ def gen_program(self):
             self.main_program[hart].label_name = label_name
             self.generate_directed_instr_stream(hart=hart,
                                                 label=self.main_program[hart].label_name,
-                                                original_instr_cnt=self.main_program[hart].instr_cnt,
+                                                original_instr_cnt=
+                                                self.main_program[hart].instr_cnt,
```
udinator, line 98: style nit:
```python
original_instr_cnt=
    self.main_program[hart].instr_cnt,
```

udinator (approved):
LGTM in general with a small nit - thanks for the effort!

I do have one general question though, especially since it seems that a large number of the constraints have been rewritten using the updates to PyVSC. Have you noticed an impact in generation performance (if any) with these updates? If so, is it something that we need to be concerned about?

aneels3:
> LGTM in general with a small nit - thanks for the effort!
> 
> I do have one general question though, especially since it seems that a large number of the constraints have been rewritten using the updates to PyVSC. Have you noticed an impact in generation performance (if any) with these updates? If so, is it something that we need to be concerned about?

Yes
There is some performance issue. 
I will discuss this in today's meeting.

hodjat91 (approved):
LGTM with the changes! Looking forward to our meeting and discuss about performance. Thanks!

PR #769, opened 2021-01-16
[Pygen]: Add multiprocessing code block for multiple iterations.

aneels3:

PR #773, opened 2021-01-28
[Pygen]: Add vector Interrupt handler

aneels3:
This PR adds the vector interrupt handler along with the riscv_signature_pkg class.
PS: Merge this after #769

### DIFF
--- a/pygen/pygen_src/riscv_asm_program_gen.py
+++ b/pygen/pygen_src/riscv_asm_program_gen.py
@@ -550,7 +550,7 @@ class riscv_asm_program_gen:
                                                       cause.name) +
                          "{}srli x{}, x{}, {}\n".format(pkg_ins.indent, cfg.gpr[0].value,
                                                         cfg.gpr[0].value, rcs.XLEN - 1) +
-                         "{}bne x{}, x0, {}{}mode_instr_handler".format(pkg_ins.indent,
+                         "{}bne x{}, x0, {}{}mode_intr_handler".format(pkg_ins.indent,
                                                                         cfg.gpr[0].value,
                                                                         pkg_ins.hart_prefix(hart),
                                                                         mode))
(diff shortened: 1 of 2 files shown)


ishita71:
@googlebot I consent

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -550,10 +552,10 @@ def gen_trap_handler_section(self, hart, mode, cause, tvec,
                                                       cause.name) +
                          "{}srli x{}, x{}, {}\n".format(pkg_ins.indent, cfg.gpr[0].value,
                                                         cfg.gpr[0].value, rcs.XLEN - 1) +
-                         "{}bne x{}, x0, {}{}mode_instr_handler".format(pkg_ins.indent,
-                                                                        cfg.gpr[0].value,
-                                                                        pkg_ins.hart_prefix(hart),
-                                                                        mode))
+                         "{}bne x{}, x0, {}{}mode_intr_handler".format(pkg_ins.indent,
+                                                                       cfg.gpr[0].value,
```
hodjat91, line 556: nit: if gpr[0] is a vsc type, maybe use get_val method instead?

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -570,10 +572,49 @@ def gen_trap_handler_section(self, hart, mode, cause, tvec,
...
 
     def gen_interrupt_vector_table(self, hart, mode, status, cause, ie,
                                    ip, scratch, instr):
-        pass
+        '''In vector mode, the BASE address is shared between interrupt 0 and exception handling.
+           When vectored interrupts are enabled, interrupt cause 0, which corresponds to user-mode
+           software interrupts, are vectored to the same location as synchronous exceptions. This
+           ambiguity does not arise in practice, since user-mode software interrupts are either
+           disabled or delegated'''
+        instr.extend((".option norvc;", "j {}{}mode_exception_handler".
```
hodjat91, line 584: nit: please move "format" to its previous line.

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -570,10 +572,49 @@ def gen_trap_handler_section(self, hart, mode, cause, tvec,
...
+                                 cfg.gpr[0].value, cfg.gpr[0].value, hex(rcs.XLEN - 1)),
+                                 "beqz x{}, 1f".format(cfg.gpr[0].value)))
+            self.gen_signature_handshake(
+                instr=intr_handler, signature_type=signature_type_t.WRITE_CSR, csr=status)
+            self.gen_signature_handshake(
+                instr=intr_handler, signature_type=signature_type_t.WRITE_CSR, csr=cause)
+            self.gen_signature_handshake(
+                instr=intr_handler, signature_type=signature_type_t.WRITE_CSR, csr=ie)
+            self.gen_signature_handshake(
+                instr=intr_handler, signature_type=signature_type_t.WRITE_CSR, csr=ip)
```
hodjat91, line 611: Could you please do these four calls to the same function in a for loop for different csrs?

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -704,9 +745,66 @@ def gen_test_file(self, test_name):
...
         logging.info("%0s is generated", test_name)
 
-    def gen_signature_handshake(self, instr, signature_type, core_status = "INITIALIZED",
-                                test_result = "TEST_FAIL", csr = "MSCRATCH", addr_label = ""):
-        pass
+    def gen_signature_handshake(self, instr, signature_type,
+                                core_status = core_status_t.INITIALIZED,
+                                test_result = test_result_t.TEST_FAIL,
+                                csr = privileged_reg_t.MSCRATCH,
+                                addr_label = ""):
```
hodjat91, line 752: nit: no space around "="

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -704,9 +745,66 @@ def gen_test_file(self, test_name):
...
+        if cfg.require_signature_addr:
+            string = []
+            string = ("li x{}, {}".format(cfg.gpr[1].value, hex(cfg.signature_addr)))
+            instr.extends(string)
```
hodjat91, line 756: Please remove s from "extends"

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -704,9 +745,66 @@ def gen_test_file(self, test_name):
...
+            # A single data word is written to the signature address.
+            # Bits [7:0] contain the signature_type of CORE_STATUS, and the upper
+            # XLEN-8 bits contain the core_status_t data.
+            if signature_type == signature_type_t.CORE_STATUS:
+                string.extend(("li x{}, {}".format(cfg.gpr[0].value, hex(core_status.value)),
+                               "slli x{}, x{}, 8".format(cfg.gpr[0].value, cfg.gpr[0].value),
+                               "addi x{}, x{}, {}".format(cfg.gpr[0].value, cfg.gpr[0].value,
+                                                          hex(signature_type.value)),
+                               "sw x{}, 0(x{})".format(cfg.gpr[0].value, cfg.gpr[1].value)))
+                instr.extend(string)
```
hodjat91, line 766: Is there a reason here for not extending the "instr" directly? I mean why would we need to extend the string first, then extend the instr?

pygen/pygen_src/riscv_asm_program_gen.py
```diff
@@ -704,9 +745,66 @@ def gen_test_file(self, test_name):
...
+            # A single data word is written to the signature address.
+            # Bits [7:0] contain the signature_type of TEST_RESULT, and the upper
+            # XLEN-8 bits contain the test_result_t data.
+            elif signature_type == test_result_t.TEST_RESULT:
+                string.extend(("li x{}, {}".format(cfg.gpr[0].value, hex(test_result.value)),
+                               "slli x{}, x{}, 8".format(cfg.gpr[0].value, cfg.gpr[0].value),
+                               "addi x{}, x{}, {}".format(cfg.gpr[0], cfg.gpr[0],
+                                                          hex(signature_type.value)),
+                               "sw x{}, 0(x{})".format(cfg.gpr[0].value, cfg.gpr[1].value)))
+                instr.extend(string)
```
hodjat91, line 776: I believe there is a tiny bug here: logically, we extend the instr once in line 756, then in the next "if else" statements, we extend the string, and again extend the instr. This way, I believe we are appending line 755 multiple times which could get problematic. Does that make sense?

pygen/pygen_src/riscv_asm_program_gen.py
aneels3: Hi @hodjat91 
Yes, its a vsc type. I tried without value it works in that case also.
I was assuming enum.name will return string name and enum.value returns an integer value.
It seems like python "{}".format() takes the integer value by default unless we specify enum.name.
Well, I have did the changes all over the asm class. 
Please take a look.

pygen/pygen_src/riscv_asm_program_gen.py
aneels3: fixed!

pygen/pygen_src/riscv_asm_program_gen.py
aneels3: thanks for pointing this.
 it's fixed now!

pygen/pygen_src/riscv_asm_program_gen.py
aneels3: yes, It makes sense now.
I have removed the string list from here and directly did the extend with instr only.

hodjat91 (approved):
LGTM!

aneels3:
Hi @udinator 
I have done the rebase.
Please go ahead with the merging such that I can rebase #779 as well.

udinator:
Thanks @aneels3!

PR #782, opened 2021-02-25
[questa] Remove -access=rwc from vlog command line arguments

### LINKED ISSUE #780: Questa simulator tool invalid vlog argument
What version of questasim is currently supported with `riscv-dv`?

When using `questa/2020.3` the simulator complains that `-access=rwc` is not a valid argument in [simulator.yaml](https://github.com/google/riscv-dv/blob/master/yaml/simulator.yaml). 
```
- tool: questa
  compile:
    cmd:
      - "vmap mtiUvm $QUESTA_HOME/questasim/uvm-1.2"
      - "vlog -64
        +incdir+<setting>
        +incdir+<user_extension>
        -access=rwc
        -f <cwd>/files.f
        -sv
        -mfcu -cuname design_cuname
        +define+UVM_REGEX_NO_DPI
        -writetoplevels <out>/top.list
        -l <out>/compile.log <cmp_opts>"
      - "vopt -64 -debug
        +designfile -f <out>/top.list
        -l <out>/optimize.log <cmp_opts>
        -o design_opt"
 ```
 
 I'm able to successfully run when removing `-access=rwc`.

rswarbrick:
This doesn't seem to be supported by at least Questa 2020.03 (see
issue #780). I don't have a Questa licence or manual locally, but it
looks like this is an Xcelium flag. Maybe it was supported by other
Questa versions as a compatibility shim? If we need to allow the
equivalent access, I think the argument is "+acc", but I'm not
convinced it's needed at all, so suggest removing it entirely.

Fixes #780.

### DIFF
--- a/yaml/simulator.yaml
+++ b/yaml/simulator.yaml
@@ -57,7 +57,6 @@
       - "vlog -64
         +incdir+<setting>
         +incdir+<user_extension>
-        -access=rwc
         -f <cwd>/files.f
         -sv
         -mfcu -cuname design_cuname


rswarbrick:
Hmm, this just failed CI, but not with a helpful message. Could someone from the riscv-dv team have a look to see what I broke (or whether it's a transient failure)?

udinator:
Hi @rswarbrick, apologies for review delay.

We've just recently set up a CI flow for riscv-dv, and it is still a little flaky on the back-end.

Could you please rebase your PR with the latest upstream? I've added a temporary fix for the CI failures.

PR #791, opened 2021-04-13
[Pygen] Add coverage support for RV32M 

aneels3:


ishita71:
@googlebot I consent

udinator (approved):
LGTM overall!
Could you please check that the coverage matches that of the SV RISCV-DV when using M extension?

aneels3:
Hi @udinator 
Sorry for the late response on it.
I have compared the results with SV coverage and pyflow separately for the same assembly file generated using pyflow.
It seems like there is some discrepancy (difference of 5-10 %) in the percentage score for the instructions.
This is observed in the RV32I extension instructions as well.

I am working on the debugging of this issue.

udinator:
@aneels3 - that's interesting...let's merge this PR but keep debugging the coverage mismatches.

PR #802, opened 2021-06-27
Update var binary of convert2bin in riscv_instr.sv

AryamanAg:
Fixed length of binary in riscv_instr.sv and riscv_compressed_instr.sv


udinator (approved):
thanks for the change!

PR #803, opened 2021-07-02
Fixed wrong length of I, S, B-type immediates

silabs-hfegran:
Previous setting caused sign extension of the second-last bit of the immediates for instructions such as JALR.


silabs-hfegran:
Perhaps I should have left out changing the python sources - the imm_len variable appears incorrect there as well, but the python code seems to fail the ci-check due to unrelated issues.

udinator (approved):
thanks for this change!

No need to worry too much about the Python CI failures, the Python codebase is still under active development and isn't completely stable yet.

Issue #778, opened 2021-02-08
init_machine_mode code potentially triggers infinite loop.


silabs-hfegran:
setup_mmode_reg(), called by enter_privileged_mode() in the gen_privileged_mode_switch_routine() enables the mstatus mie-flag and enables interrupts before mret is called. If an interrupt request is present when this flag is set, an infinite loop will occur:

Take the following example of generated code:
```
init_machine_mode:
                  li x26, 0x18ba
                  csrw 0x300, x26 # MSTATUS
                  li x26, 0xb0b
                  csrw 0x304, x26 # MIE
                  mret
```
If an external interrupt is triggered at the mret instruction above, the return address for the trap (mepc) will be set to this mret. This will cause the final mret to return to itself, and end up in an infinite loop, as no further updates to mepc is performed.

From the Risc-V privileged spec:
> When executing an xRET instruction, supposing xPP holds the value y, xIE
is set to xPIE; the privilege mode is changed to y; xPIE is set to 1; and xPP is set to U (or M if
user-mode is not supported).

Perhaps it is safer to not enable MIE directly, only MPIE in this code section, and let mret take care of enabling the exception when executing mret.

PR #804, opened 2021-07-02
Issue #778 fix, change mie behavior in setup_mmode_reg

silabs-hfegran:
MIE is explicitly set before mret is called, leading to potential infinite loop in case of an interrupt between interrupt enable and mret. Instead taken care of by MPIE as per Riscv-spec (Fix for issue #778 )


udinator (approved):
This change makes sense to me, seems to be a nice clean solution.
Thanks for implementing it!

PR #811, opened 2021-07-20
[Pygen]: Add gen_timeout for PyFlow

aneels3:
Signed-off-by: aneels3 <b150023ec@nitsikkim.ac.in>

### DIFF
--- a/run.py
+++ b/run.py
@@ -380,11 +380,21 @@ def gen(test_list, argv, output_dir, cwd):
     # Run the instruction generator
     if not argv.co:
         seed_gen = SeedGen(argv.start_seed, argv.seed, argv.seed_yaml)
+        if argv.simulator == 'pyflow':
+            """Default timeout of Pyflow is 20 minutes, if the user
+               doesn't specified their own gen_timeout value from CMD
+            """
+            if argv.gen_timeout == 360:
+                gen_timeout = 1200
+            else:
+                gen_timeout = argv.gen_timeout
+        else:
+            gen_timeout = argv.gen_timeout
         do_simulate(sim_cmd, argv.simulator, test_list, cwd, argv.sim_opts,
                     seed_gen,
                     argv.csr_yaml, argv.isa, argv.end_signature_addr,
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -380,11 +380,18 @@ def gen(test_list, argv, output_dir, cwd):
     # Run the instruction generator
     if not argv.co:
         seed_gen = SeedGen(argv.start_seed, argv.seed, argv.seed_yaml)
+        if argv.simulator == 'pyflow':
```
udinator, line 383: I feel this 20min timeout for PyFlow should be enforced by default (and add a comment explaining there are some performance issues).

then we can simplify this to:
```python
gen_timeout = 1200 if argv.simulator == 'pyflow' else argv.gen_timeout
```

wdyt?

aneels3: I will add the comment over there.

For the second part,
If the user tries to run more than 10k instructions with multiple iterations (say >20) and also provides the own value for --gen_timeout through the CLI.
Then it should take the user-specified gen_timeout value.

WDYT on this?
```python3
gen_timeout = 1200 if argv.simulator == 'pyflow' else argv.gen_timeout
```
The above-simplified logic will restrict users from specifying their own gen_timeout in case of pyflow

hodjat91: I agree with Anil. Setting 20 minutes as a default timeout value for PyFlow makes sense. However, it's good to also give the user the freedom of picking their desired timeout, given the fact that for larger programs, PyFlow could take more than 20 minutes. What do you think @udinator ?

hodjat91 (approved):

run.py
udinator: Good point, LGTM.

@aneels3 can you add a comment saying that the default timeout of PyFlow is 20min if the user doesn't specify their own values from cmdline?

udinator (approved):

run.py
aneels3: Done!

aneels3: Thanks @hodjat91  and @udinator

PR #832, opened 2021-11-16
Bitmanip v1.0.0 support

silabs-hfegran:
This commits add support for the Zba, Zbb, Zbc and Zbs extensions. 

`define BITMANIP_V1_0_0 in riscv_instr_pkg.sv enables these added features,
and support for 0.92 has been retained, selectable by instead defining BITMANIP_V0_9_2

Similar to the previous b-extension, the zb*-extensions are enabled by setting the +enable_zb*_extension plusargs and adding the required isas (RV32ZB* / RV64ZB*) to the list of supported isa.


weicaiyang (approved):
LGTM. Thanks for adding this support.
@tomroberts-lowrisc can you also take a look and if this works for Ibex?

tomroberts-lowrisc:
@vogelpi @GregAC

weicaiyang:
> @vogelpi @GregAC

Sorry, Tom. I actually wanted to ping Greg to take a look...

vogelpi:
Thanks for this PR @silabs-hfegran, this is great stuff!

I've successfully tested it with Ibex which is still on v.0.92 but right now also has Zba from v.1.0.0. Both modes work and so does the switch using the define inside `riscv_instr_pkg.sv`. Since v.1.0.0 only comprises Zba, Zbb, Zbc and Zbs and we need to keep supporting the other sub-extensions such as Zbt, Zbp, and Zbr that are still part of v.0.9.4 we might want to add support for mixing v.1.0.0 and v.0.9.4 whereas the switch currently supports the version exclusively. But this probably requires some more thought and discussion and best goes into a separate PR. So this LGTM!

silabs-hfegran:
Thanks for the feedback! I think supporting both 0.94 and 1.0.0 should be fairly easy to accomplish - I opted not to do any changes to the existing infrastructure to avoid breaking backwards compatibility for those who depend on it. If support for 0.92 is no longer needed then the change would largely just involve removing the duplicate instructions from the old B-extension and updating the ones that are outdated or missing. I do feel however, that those also would benefit from being split into their own submodules.

vogelpi:
Thanks for getting back @silabs-hfegran , I agree that ultimately support for 0.92 will no longer be needed. But right now we are in a awkward transition phase a need to update DV, toolchain, and of course the RTL. I think my next step is to align the Zbb and Zbs RTL with 1.00. After that I will probably follow your suggested approach for the remaining sub-extensions incl. splitting into submodules.

weicaiyang:
Thanks @silabs-hfegran and @vogelpi. Let me merge this PR.

PR #837, opened 2021-12-08
[style] Break long lines in newly added files

msfschaffner:
These >100 char lines cause issues in OpenTitan when vendoring in Ibex due to our Verible style linter.

Signed-off-by: Michael Schaffner <msf@google.com>


weicaiyang (approved):
thanks for fixing this styles

PR #839, opened 2021-12-17
Fix broken run.py script

puneet:
Delimiter was missing following the seed argument that was added in
run.py a couple of commits back.


puneet:
I have uploaded the CLA. Please rescan.

hodjat91 (approved):
LGTM!

PR #843, opened 2022-01-21
Add register definitions for privilege spec 1.12 and debug spec 1.0.0

silabs-hfegran:
Updated CSR register enum with registers defined in privilege spec v1.12 and debug spec v1.0.0


silabs-hfegran:
@weicaiyang, it would be great if you could have a look at this.

weicaiyang:
> @weicaiyang, it would be great if you could have a look at this.

I'm not familiar with this part. perhaps @hodjat91 can take a look.

BTW, looks like only 1st commit is related. Can you remove the other 2 commits in this PR?

silabs-hfegran:
> BTW, looks like only 1st commit is related. Can you remove the other 2 commits in this PR?

Thanks, that should be fixed now

hodjat91:
> > @weicaiyang, it would be great if you could have a look at this.
> 
> I'm not familiar with this part. perhaps @hodjat91 can take a look.
> 
> BTW, looks like only 1st commit is related. Can you remove the other 2 commits in this PR?

Apologies for the delay. Please add me as a reviewer next time, so I would get notified.

hodjat91 (approved):
LGTM and thanks!

PR #844, opened 2022-01-26
lib.py, launch process in new session to fix timeout issue

yanicasa:
I propose a solution for a problem I found concerning the management of jobs that reach the timeout threshold.
Indeed, if the job creates child processes, they are not killed by the current management.

According to python docs, the fix works only on linux environment, It should works as before on windows thanks to try/except

### DIFF
--- a/scripts/lib.py
+++ b/scripts/lib.py
@@ -24,6 +24,7 @@ import subprocess
 import time
 import yaml
 import logging
+import signal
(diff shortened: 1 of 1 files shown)


mcockrell-google (approved):

hodjat91 (approved):
LGTM, thanks!

PR #849, opened 2022-02-17
Allow ISA extensions in parse_iss_yaml

rswarbrick:
At lowRISC, we're using `rv32imc_Zba_Zbb_Zbc_Zbs_Xbitmanip` as our ISA
string. This has xlen "32", variant "imc" and then some other flags
denoting extensions.

The existing code worked just fine for Spike (the ISS we're most
interested in), because the YAML entry for Spike doesn't actually use
the '`<xlen>`' or '`<variant>`' variables. But it printed out an "Illegal
ISA" error message, which is a bit confusing.

Fix that, by allowing (and ignoring) ISA extension flags.


rswarbrick:
Ping! Is anyone maintaining this project any more?

PR #863, opened 2022-05-24
[epmp] Add support for mseccfg CSR

vogelpi:
This PR adds basic support for the mseccfg CSR required for ePMP. Similar to the actual PMP configuration:
- the desired configuration for mseccfg can be provided using an argument, e.g. `+mseccfg=MML:0,MMWP:0,RLB:1`.
- if randomization is turned on, it's still possible to force the bits of `mseccfg` partially or completely using this argument.
 
The CSR is only configured if the bit `support_epmp` in `riscv_core_setting.sv` is set to 1.


vogelpi:
Hi @hodjat91 and @weicaiyang ,
would you mind taking a look at this PR? We need this for Ibex in OpenTitan. Thanks!

hodjat91 (approved):
LGTM!

vogelpi:
Thanks a lot @hodjat91 for taking a look and merging this.

Issue #819, opened 2021-09-03
Issue in operands of generated compressed float load-store instructions


UrvishkumarPatel:
I generated an asm file using test "riscv_floating_point_rand_test" in target "riscv-dv/target/rv32imafdc"

But while compiling I got the following error messages:

> asm/riscv_floating_point_rand_test_0.S: Assembler messages:
> asm/riscv_floating_point_rand_test_0.S:5200: Error: illegal operands `c.fsw a0,68(a1)'

It seems that the programs are having issue in compressed float load stores. The operands taken must be floating point registers and not integer register.

PR #864, opened 2022-05-25
Compressed floating point  load-store instruction to have FPR as operands.

aneels3:
Signed-off-by: aneels3 <b150023ec@nitsikkim.ac.in>

This PR fixes the #819 issue related to compressed floating-point instructions which require floating-point registers as operands.
Changes are being made for both Single and Double floating-point load and store instructions.


aneels3:
Hi @hodjat91 and @weicaiyang 
Could you review the changes made?

PR #873, opened 2022-07-14
[Smepmp] Fixes `gen_pmp_instr` when MML and MMWP are enabled

marnovandermaas:
The problem that I was running into with machine mode lockdown (MML) and machine mode whitelist policy (MMWP) is that I needed the instructions under the `pmp_setup` label (eventually in PMP region 0) to be set to executable. Otherwise it would set `MSECCFG` and then trap on the instruction immediately after. However, we still want to set `MSECCFG` before configuring the PMP configs because some PMP setups are valid depending on the Smepmp mode.

My proposed solution is to change the `gen_pmp_instr` function to take these steps:
- Set address of region 0 before setting MSECCFG
- If  MML, initially set MSECCFG to MML=0, MMWP=0, RLB=1
- If  MML, set the config of region 0 to LXWR=1100 and TOR
- If MMWP, set the config of region 0 to LXWR=0111 and TOR
- If MML or MMWP, set requested MSECCFG with RLB hardcoded to 1
- Don't override region 0 config if `+pmp_region_0` is passed
- Change default region 0 config in MML mode to shared execute
- Set all other addresses and configs as before
- Set requested MSECCFG (including RLB)

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -292,42 +292,91 @@ class riscv_pmp_cfg extends uvm_object;
   // Since either 4 (in rv32) or 8 (in rv64) PMP configuration registers fit into one physical
   // CSR, this function waits until it has reached this maximum to write to the physical CSR to
   // save some extraneous instructions from being performed.
+  //
+  // The general flow of this function:
+  // - Set address of region 0 before setting MSECCFG
+  // - If  MML, initially set MSECCFG to MML=0, MMWP=0, RLB=1
+  // - If  MML, set the config of region 0 to LXWR=1100 and TOR
+  // - If MMWP, set the config of region 0 to LXWR=0100 and TOR
+  // - If MML or MMWP, set requested MSECCFG with RLB hardcoded to 1
+  // - Don't override region 0 config if `+pmp_region_0` is passed
+  // - Set default region 0 config in MML mode to shared execute
+  // - Set all other addresses and configs
+  // - Set requested MSECCFG (including RLB)
   function void gen_pmp_instr(riscv_reg_t scratch_reg[2], ref string instr[$]);
     bit [XLEN - 1 : 0] pmp_word;
     bit [XLEN - 1 : 0] cfg_bitmask;
     bit [7 : 0] cfg_byte;
     int pmp_id;
+    string arg_value;
+
+    // Load the address of the <main> section into pmpaddr0.
+    // This needs to be done before setting MSECCFG in the case of MML or MMWP.
+    instr.push_back($sformatf("la x%0d, main", scratch_reg[0]));
+    instr.push_back($sformatf("srli x%0d, x%0d, 2", scratch_reg[0], scratch_reg[0]));
+    instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmp_addr, scratch_reg[0]));
+    `uvm_info(`gfn, "Loaded the address of <main> section into pmpaddr0", UVM_LOW)
+
     if (riscv_instr_pkg::support_epmp) begin
-      `uvm_info(`gfn, $sformatf("MSECCFG: MML %0x, MMWP %0x, RLB %0x", mseccfg.mml,
-          mseccfg.mmwp, mseccfg.rlb), UVM_LOW)
-      cfg_byte = {mseccfg.rlb, mseccfg.mmwp, mseccfg.mml};
+      // In case of MML or MMWP we need to set region 0 to executable before setting MSECCFG.
+      if (mseccfg.mml == 1 || mseccfg.mmwp == 1) begin
+        if (mseccfg.mml == 1) begin
+          // Writing MSECCFG with RLB set to 1 to stop the config with L enabled from locking
+          // everything before configuration is done.
+          `uvm_info(`gfn, $sformatf("MSECCFG: MML 0, MMWP 0, RLB 1"), UVM_LOW)
+          cfg_byte = {1'b1, 1'b0, 1'b0};
+          instr.push_back($sformatf("csrwi 0x%0x, %0d", MSECCFG, cfg_byte));
+          // This configuration needs to be executable in M-mode both before and after writing to
+          // MSECCFG. It will deny execution for U-Mode, but this is necessary because RWX=111 in
+          // MML means read only, and RW=01 is not allowed before MML is enabled.
+          cfg_byte = {1'b1, pmp_cfg[0].zero, TOR, 1'b1, 1'b0, 1'b0};
+        end else begin
+          // We must set pmp region 0 to executable before enabling MMWP. RW=00 to be consistend
+          // with MML configuration as much as possible.
+          cfg_byte = {1'b0, pmp_cfg[0].zero, TOR, 1'b1, 1'b0, 1'b0};
+        end
+        // Enable the selected config on region 0.
+        `uvm_info(`gfn, $sformatf("temporary pmp_word: 0x%0x", cfg_byte), UVM_DEBUG)
+        instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg[0], cfg_byte));
+        instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmpcfg_addr, scratch_reg[0]));
+      end
+      // Writing MSECCFG with RLB still set to 1 otherwise we cannot complete configuration.
+      `uvm_info(`gfn, $sformatf("MSECCFG: MML %0x, MMWP %0x, RLB 1", mseccfg.mml, mseccfg.mmwp),
+                UVM_LOW)
+      cfg_byte = {1'b1, mseccfg.mmwp, mseccfg.mml};
       instr.push_back($sformatf("csrwi 0x%0x, %0d", MSECCFG, cfg_byte));
     end
+
     foreach (pmp_cfg[i]) begin
-      // TODO(udinator) condense this calculations if possible
       pmp_id = i / cfg_per_csr;
-      if (i == 0) begin
-        cfg_byte = {1'b0, pmp_cfg[i].zero, TOR, 1'b1, 1'b1, 1'b1};
+      // In case an argument is give, we want to allow the test writer to change region 0 to
+      // anything they like even if it means that a trap occurs on the next instruction. If no
+      // argument is given, we should set region 0 to have a configuration with execute enabled.
+      if (i == 0 && !inst.get_arg_value("+pmp_region_0=", arg_value)) begin
+        if(riscv_instr_pkg::support_epmp && mseccfg.mml) begin
+          // This value is different from above (M-mode execute only) because we need region 0 to
+          // be executable in both M-mode and U-mode, since RISCV-DV switches privledge before
+          // <main> but after <pmp_setup>. We choose not to use the shared code region that also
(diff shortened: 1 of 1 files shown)


src/riscv_pmp_cfg.sv
```diff
@@ -371,6 +370,12 @@ class riscv_pmp_cfg extends uvm_object;
         pmp_word = 0;
       end
     end
+    if (riscv_instr_pkg::support_epmp) begin
+      `uvm_info(`gfn, $sformatf("MSECCFG: MML %0x, MMWP %0x, RLB %0x", mseccfg.mml,
+          mseccfg.mmwp, mseccfg.rlb), UVM_LOW)
+      cfg_byte = {mseccfg.rlb, mseccfg.mmwp, mseccfg.mml};
+      instr.push_back($sformatf("csrwi 0x%0x, %0d", MSECCFG, cfg_byte));
+    end
```
vogelpi, line 378: Hey @marnovandermaas ,
I thought about configuring `mseccfg` last as well. I am not sure if this will work because depending on the `mseccfg`, some PMP configuration become legal which are illegal otherwise. That's why I've put the write to `mseccfg` first.

marnovandermaas: @vogelpi, the problem that I was running into with machine mode lockdown (MML) and machine mode whitelist policy (MMWP) is that I needed the PMP regions to be set up correctly before enabling those modes. Specifically, I needed the instructions under the `pmp_setup` label (eventually in PMP region 0) to be set to executable. Otherwise it would set `MSECCFG` and then trap on the instruction immediately after. I thought the best solution would be to flip the order: set the PMP regions up and only after limit M-mode's access. This does mean that if we ever configure PMP region 0 with `X` set to zero that we will cause an immediate trap.

As an alternative I could write PMP region 0 first in case MML or MMWP are selected and configure all other regions afterwards, but I'm not sure that has any benefits over the change I am currently proposing. What do you think?

marnovandermaas: One problem is that if the PMP is not in MML mode, the values with `R=0` and `W=1` are reserved, while these values are used if MML mode is enabled. So maybe the approach of setting up PMP region 0 as being executable before writing to `MSECCFG` and then setting the other regions up afterwards is better.

vogelpi: Thanks @marnovandermaas  for explaining the situation. What I had in mind is this (see Smepmp spec, Chapter 2, Point 4, b):
> ... When When mseccfg.MML is set the system’s behavior changes in the following way:
b) Adding a new M-mode-only or a locked Shared-Region rule with executable privileges is not possible
and such pmpcfg writes are ignored, leaving pmpcfg unchanged. This restriction can be temporarily
lifted e.g. during the boot process, by setting mseccfg.RLB.

I think implementing the alternative solution you suggested would actually be better. Because this would allow the verification of both the scenario you describe and what I had in mind originally. I think the last part
> This restriction can be temporarily lifted e.g. during the boot process, by setting mseccfg.RLB.

Can just be covered by randomizing RLB so that in some runs the restriction is bypassed and in some runs not. But we should probably add cover points for this.

vogelpi: Our answers collided but it seems we are both thinking the same. Your plan sounds good to me. Let me know if you want me to have another look or if you want to discuss.

marnovandermaas: @vogelpi do you think my latest update fixes what we talked about? I generate a few extra instructions, but I think the result is that we can test more legal configurations of the Smepmp.

src/riscv_pmp_cfg.sv
```diff
@@ -297,17 +297,49 @@ class riscv_pmp_cfg extends uvm_object;
...
+      // In case of MML or MMWP we need to set region 0 to executable before setting MSECCFG
+      if (mseccfg.mml != 0 || mseccfg.mmwp != 0) begin
+        if (mseccfg.mml != 0) begin
+          // Writing MSECCFG with RLB set to 1 to stop the pmp write from locking everything
+          `uvm_info(`gfn, $sformatf("MSECCFG: MML 0, MMWP 0, RLB 1"), UVM_LOW)
+          cfg_byte = {1'b1, 1'b0, 1'b0};
+          instr.push_back($sformatf("csrwi 0x%0x, %0d", MSECCFG, cfg_byte));
+          cfg_byte = {1'b1, pmp_cfg[0].zero, TOR, 1'b1, 1'b0, 1'b0};
+        end else begin
+          cfg_byte = {1'b0, pmp_cfg[0].zero, TOR, 1'b1, 1'b1, 1'b1};
```
vogelpi, line 318: I think I don't understand this. We're setting RWX for Region 0 and afterwards, MMWP will be enabled. According to Page 4 in the Smepmp spec, does this mean any M mode access to region 0 will be allowed afterwards? And any accesses outside Region 0 will be denied? Can you please explain to me what we are doing here and why? Maybe a comment would be useful.

src/riscv_pmp_cfg.sv
```diff
@@ -297,17 +297,49 @@ class riscv_pmp_cfg extends uvm_object;
...
-          mseccfg.mmwp, mseccfg.rlb), UVM_LOW)
-      cfg_byte = {mseccfg.rlb, mseccfg.mmwp, mseccfg.mml};
+      // In case of MML or MMWP we need to set region 0 to executable before setting MSECCFG
+      if (mseccfg.mml != 0 || mseccfg.mmwp != 0) begin
+        if (mseccfg.mml != 0) begin
+          // Writing MSECCFG with RLB set to 1 to stop the pmp write from locking everything
+          `uvm_info(`gfn, $sformatf("MSECCFG: MML 0, MMWP 0, RLB 1"), UVM_LOW)
+          cfg_byte = {1'b1, 1'b0, 1'b0};
+          instr.push_back($sformatf("csrwi 0x%0x, %0d", MSECCFG, cfg_byte));
+          cfg_byte = {1'b1, pmp_cfg[0].zero, TOR, 1'b1, 1'b0, 1'b0};
```
vogelpi, line 316: Here we set LX, so U mode is denied, M-mode is enforced if I understand correctly.

src/riscv_pmp_cfg.sv
```diff
@@ -297,17 +297,49 @@ class riscv_pmp_cfg extends uvm_object;
...
       // TODO(udinator) condense this calculations if possible
       pmp_id = i / cfg_per_csr;
-      if (i == 0) begin
-        cfg_byte = {1'b0, pmp_cfg[i].zero, TOR, 1'b1, 1'b1, 1'b1};
+      if (i == 0 && !inst.get_arg_value("+pmp_region_0=", tmp_value)) begin
+        if(riscv_instr_pkg::support_epmp && mseccfg.mml) begin
+          cfg_byte = {1'b1, pmp_cfg[i].zero, TOR, 1'b1, 1'b1, 1'b0};
+        end else begin
+          cfg_byte = {1'b0, pmp_cfg[i].zero, TOR, 1'b1, 1'b1, 1'b1};
+        end
```
vogelpi, line 342: I think this needs some explanation. If I understand correctly, you want to avoid that if a `pmp_region_0` is provided as argument, the permissions are overwritten. With the other changes below it seems to be that the address part of the the `pmp_region_0` argument is now completely ignored. Is this intentional? Wouldn't it be better to set everything up first with the default Region 0 set above and then finally set the Region 0 configured via argument if provided?

Also, it seems that you're changing the access rights here. Above you would set LX if MML = 1 or RWX with MML = 0. But here you set LWX if MML = 1 (RWX is the same if MML = 0). Is this intentional?

Maybe, it's just a matter of putting some more comments.

vogelpi (commented):
Hey @marnovandermaas,
thanks for reworking this. I did have another look but have some questions. I am not sure if I fully understand everything, maybe it's just a matter of adding more comments. Let me know if you prefer to have a quick call for a discussion.

src/riscv_pmp_cfg.sv
marnovandermaas: Change MMWP `cfg_byte` to be just execute, not write and read.

src/riscv_pmp_cfg.sv
marnovandermaas: Find out if we need U-mode execution of region 0 (for example if RISCV-DV generates and `mret` in this region for U-mode tests).

src/riscv_pmp_cfg.sv
```diff
@@ -297,17 +297,49 @@ class riscv_pmp_cfg extends uvm_object;
     bit [XLEN - 1 : 0] cfg_bitmask;
     bit [7 : 0] cfg_byte;
     int pmp_id;
+    string tmp_value;
+
+    // load the address of the <main> section into pmpaddr0
```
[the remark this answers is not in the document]
marnovandermaas: Thanks, should be solved now

src/riscv_pmp_cfg.sv
```diff
@@ -297,17 +297,49 @@ class riscv_pmp_cfg extends uvm_object;
...
+    instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmp_addr, scratch_reg[0]));
+    `uvm_info(`gfn, "Loaded the address of <main> section into pmpaddr0", UVM_LOW)
+
     if (riscv_instr_pkg::support_epmp) begin
-      `uvm_info(`gfn, $sformatf("MSECCFG: MML %0x, MMWP %0x, RLB %0x", mseccfg.mml,
-          mseccfg.mmwp, mseccfg.rlb), UVM_LOW)
-      cfg_byte = {mseccfg.rlb, mseccfg.mmwp, mseccfg.mml};
+      // In case of MML or MMWP we need to set region 0 to executable before setting MSECCFG
+      if (mseccfg.mml != 0 || mseccfg.mmwp != 0) begin
+        if (mseccfg.mml != 0) begin
```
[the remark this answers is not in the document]
marnovandermaas: Thanks, should be solved now.

src/riscv_pmp_cfg.sv
marnovandermaas: I think we clarified this during our call. I changed the MMWP config to be execute only to match that of MML, and I added some comments explaining this.

src/riscv_pmp_cfg.sv
marnovandermaas: Yes, we need to deny U-mode because before MML is enabled writing any value with WR=10 is not valid. I need a config that is executable both before and after MML is enabled. I've added a comment explaining this in the code.

src/riscv_pmp_cfg.sv
marnovandermaas: I've looked through the RISCV-DV code and it seems the `mret` instruction does occur before main (so while executing in region 0). I made the MML config a shared execute-only region with LXWR=1010. I've also added a comment explaining this.

src/riscv_pmp_cfg.sv
vogelpi: Yes, I think so, thanks @marnovandermaas .

vogelpi (approved):
Thanks a lot @marnovandermaas for reworking this and adding the additional comments, they are really useful. This looks good to me.

@weicaiyang , would you mind having a look as well and in case you agree to merge this?

weicaiyang (approved):

vogelpi:
Thanks @weicaiyang !

PR #877, opened 2022-07-26
Allow for WFI in User Mode

ctopal:
This commit allows the case because if mstatus.tw is low, WFI behaviour
will not change regarding the mode. If it's set to one, it will act like
an illegal instruction. That is also not a problem because we already
generate illegal instruction handler inside gen_all_trap_handler method.

Signed-off-by: Canberk Topal <ctopal@lowrisc.org>

### DIFF
--- a/pygen/pygen_src/riscv_instr_gen_config.py
+++ b/pygen/pygen_src/riscv_instr_gen_config.py
@@ -475,9 +475,6 @@ class riscv_instr_gen_config:
         logging.info("min_stack_len_per_program value = {}"
                      .format(self.min_stack_len_per_program))
         self.check_setting()  # check if the setting is legal
-        # WFI is not supported in umode
-        if self.init_privileged_mode == privileged_mode_t.USER_MODE:
-            self.no_wfi = 1
 
     def check_setting(self):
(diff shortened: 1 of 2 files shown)


ctopal:
> ... if mstatus.tw is low, WFI behaviour will not change regarding the mode. 

I think I need to clarify this a bit further. In riscv-privileged spec v20211203: WFI behaviour specified as below:
```
WFI is available in all privileged modes, and optionally available to U-mode.
This instruction may raise an illegal instruction exception when TW=1 in mstatus, as described
in Section 3.1.6.5.
```
So in the case of TW=0 in mstatus, it is either going to go to Sleep mode without a problem or handled as `illegal_instr` in the case of it not being implemented. So, there is no reason to not allow it in User mode.

MikeOpenHWGroup:
> there is no reason to not allow it in User mode.

Agree.  Also, when using riscv-dv for RTL verification, it is important to be able to generate WFI in both cases (mstatus.tw == 0,1).

GregAC (approved):

GregAC:

weicaiyang (approved):

PR #881, opened 2022-07-27
Gen csr improvements

GregAC:
There are other methods to approach this but I felt this would provide the maximum flexibility without over-complicating the CSR YAML format.

### DIFF
--- a/scripts/gen_csr_test.py
+++ b/scripts/gen_csr_test.py
@@ -48,6 +48,18 @@ TEST_RESULT = 1
 TEST_PASS = 0
 TEST_FAIL = 1
 
+class SimpleWriteCSRField:
+    def __init__(self, mask_bitarray, pos, read_only):
+        self.mask_bitarray = mask_bitarray
+        self.pos = pos
+        self.read_only = read_only
+
+    def write(self, new_field_val, csr_val):
+        # only write if not read only
+        if not self.read_only:
+            val_slice = new_field_val[self.pos[0]:self.pos[1] + 1]
+            csr_val.overwrite(self.mask_bitarray & val_slice, self.pos[0])
+
 
 def get_csr_map(csr_file, xlen):
     """
@@ -59,7 +71,7 @@ def get_csr_map(csr_file, xlen):
 
     Returns:
       A dictionary contining mappings for each CSR, of the form:
-      { csr_name : [csr_address, csr_val_bitarray, csr_write_mask_bitarray, csr_read_mask_bitarray] }
+      { csr_name : [csr_address, csr_val_bitarray, csr_write_fields, csr_read_mask_bitarray] }
     """
     rv_string = "rv{}".format(str(xlen))
     csrs = {}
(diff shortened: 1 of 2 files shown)


GregAC:
@vogelpi please take a look

scripts/gen_csr_test.py
```diff
@@ -60,6 +60,35 @@ def write(self, new_field_val, csr_val):
             val_slice = new_field_val[self.pos[0]:self.pos[1] + 1]
             csr_val.overwrite(self.mask_bitarray & val_slice, self.pos[0])
 
+class LegalizeWriteCSRField:
+    def __init__(self, mask_bitarray, pos, read_only, legalize_py):
+        self.mask_bitarray = mask_bitarray
+        self.pos = pos
+        self.read_only = read_only
+        self.legalize_py = legalize_py
+        print(self.legalize_py)
```
vogelpi, line 69: Should this `print()` be removed?

scripts/gen_csr_test.py
```diff
@@ -60,6 +60,35 @@ def write(self, new_field_val, csr_val):
...
+
+    def _legalize_val(self, orig_val):
+        legalize_globals = {
+                'val_in' : orig_val,
+                'val_out' : orig_val
+        }
+
+        exec(self.legalize_py, legalize_globals)
+
+        return legalize_globals['val_out']
```
vogelpi, line 90: Are `val_in/out` some kind of "global" variables to be used just for `legalize_py`? I am just asking out of interest, I have never seen this before.

vogelpi (approved):
Thanks @GregAC . I don't fully understand all of the involved magic around the `legalize_py` functions but this looks good.

This is needed to correctly deal with WARL registers of which there can be many. Also the refactoring of gen_csr_test looks sensible to me.

scripts/gen_csr_test.py
silabs-hfegran: This is awesome @GregAC - I was looking into a fix for the very same issue.

scripts/gen_csr_test.py
GregAC: Whoops, yes it should

scripts/gen_csr_test.py
GregAC: > Are `val_in/out` some kind of "global" variables to be used just for `legalize_py`?

Yes exactly. The `exec` call effectively runs the python in the `self.legalize_py` string in it's own interpreter environment. This sets up `val_in` and `val_out` as two global variables for it to use.

GregAC:

silabs-hfegran:
One question - from my impression of this code, this legalizes the write-values to the registers, would it be difficult to expand this to actually do the same for the expected response values?

To make an example: write b'01 to a bitfield that accepts only b'11 or b'00, 
In this case we could (depending on cpu behavior) read a value where there is no change to the bitfields as the written value is not a legal value.

GregAC:
> One question - from my impression of this code, this legalizes the write-values to the registers, would it be difficult to expand this to actually do the same for the expected response values?

It already does this. The legalized value is used by gen_csr_test to predict the next read value of the CSR.

However the python fragment has no access to the old field value and it should do as according to the RISC-V spec:

> Implementations can return any legal value on the read of a WARL field when the last write was of
an illegal value, but the legal value returned should deterministically depend on the illegal written
value and the architectural state of the hart.

Really it should have access to all other field and CSR values too. For now I think we should keep it simple and I'll add something that gives the python fragment access to the old field value. Should anyone need something more sophisticated they can open a new PR to add it.

silabs-hfegran (approved):

weicaiyang (approved):

PR #882, opened 2022-07-28
PMP configuration randomizing fixes

ctopal:
This commit fixes the incorrect assumptions about TOR mode not
allowing overlapping addresses and `PMPGranularity=0` causing
not being able to enable NAPOT mode. Both of them are incorrect
behaviours according to privileged spec v20211203.

Also cleans some whitespaces in `.d` code.

Signed-off-by: Canberk Topal <ctopal@lowrisc.org>

### DIFF
--- a/euvm/riscv/gen/riscv_pmp_cfg.d
+++ b/euvm/riscv/gen/riscv_pmp_cfg.d
@@ -30,9 +30,9 @@ import esdl.rand: rand, constraint;
 import uvm;
 
 class riscv_pmp_cfg: uvm_object {
-  
+
   mixin uvm_object_utils;
- 
+
   // default to a single PMP region
   @rand @UVM_DEFAULT int pmp_num_regions = 1;
 
@@ -83,7 +83,7 @@ class riscv_pmp_cfg: uvm_object {
   // constraints - apply when pmp_randomize is 1 //
   /////////////////////////////////////////////////
 
-  
+
   constraint! q{
     pmp_num_regions inside [1 : 16];
     pmp_granularity inside [0 : XLEN + 3];
@@ -98,7 +98,6 @@ class riscv_pmp_cfg: uvm_object {
 
   constraint! q{
     foreach (cfg; pmp_cfg) {
-      (pmp_granularity == 0) ->	(cfg.a != pmp_addr_mode_t.NAPOT);
       (pmp_granularity >= 1) -> (cfg.a != pmp_addr_mode_t.NA4);
     }
   } grain_addr_mode_c;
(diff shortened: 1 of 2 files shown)


ctopal:
More clarity around false constraining:

In privileged spec V20211203 p.g. 59 you have the following note 

```
If PMPADDRi−1 ≥ PMPADDRi and PMPCFGi.A=TOR, then PMP entry i matches no addresses
```
Which suggests an option for having overlapping addresses for TOR mode.

About PMPGranularity=0 means no NAPOT mode one: There is no part in the spec that disallows that. Only a mention about PMPGranularity>0 causing NA4 to be unselectable (which makes sense because that guarantees smallest possible PMP grain to be greater than 4 bytes.)

ctopal:
Another note about ePMP and how it changes things:

In Smepmp Version 1.0, 12/2021 it specifies a PMP rule as such:

```
 PMP Rule: The contents of a pmpcfg register and its associated pmpaddr register(s),
that encode a valid protected physical memory region, where pmpcfg[i].A != OFF, and
if pmpcfg[i].A == TOR, pmpaddr[i-1] < pmpaddr[i].
```

Which suggests the constraint in the code was accurate in the first place. This is really confusing to me, should I just make the change in a way that only allows for that constraint to exists if epmp is enabled? Would love some suggestions about it. @GregAC, WDYT?

marnovandermaas:
My interpretation of the Smepmp spec is that you can still configure the PMP in such a way, but that the region will not be valid and therefore no addresses will match it.

GregAC:
For the overlapping address what's happening is without `pmp_allow_addr_overlap` being set addresses aren't constrained relative to each other so in particular you could end up with a TOR region that isn't configured in a valid way (the lower address CSR holds the higher address bound). Here the easy way out was taken by just disabling all overlapping regions to ensure you always ended up with valid TOR regions. Two points to make here:

1. You want to generated invalid things to ensure they behave as expected (in this case produce a region that never matches as @marnovandermaas says)
2. Even if you want only valid regions you can still generate overlapping TOR regions

Totally removing the constraint that disables overlapping regions if we have a TOR region is fine, though we will end up with a high frequency of useless/invalid TOR regions.

I think we should add a new constraint that checks a regions address against it's lower neighbour if that region is a TOR region and we've set some 'no_invalid_tor_regions' bit.

marnovandermaas:
@ctopal and I talked about how to fix this and make the following proposal:
- For NA4 generate a random address greater than `_start`
- For NAPOT generate a random range and then randomize the top bits. Example piece of code `$urandom() || !(1 << $urandom(0,31)`. These ranges should include addresses greater than `_start`.
- For TOR choose a random offset and calculate the address based off of the address of the previous entry plus the offset

For all modes, instead of having a flag `+pmp_allow_addr_overlap`, we should create a flag that says `+allow_full_random_address` which allows all addresses to be anything. For example, before or after `_start` and TOR will be able to create regions where the address is less than the previous.

GregAC:
@ctopal what became of this work? Has it been added elsewhere or do we still need to merge this?

ctopal:
I think I'm just going to close this one since Marno's PR actually does the same change as well. https://github.com/google/riscv-dv/pull/903/commits/64132dcf28c2055c5a5cfb4edcea7fc3cba69843. We don't need two of them merged.

PR #885, opened 2022-08-02
[PMP] Randomizing entry for instructions for PMP randomization

marnovandermaas:
The problem with the original design was that region 0 would always be dedicated to be TOR and from the base of DRAM to `main`. This caused us to not hit all the coverage points that we wanted: https://github.com/lowRISC/ibex/issues/1702
This PR fixes this by generating a random code entry, so that region 0 also gets exercised when `pmp_randomize` is enabled.

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -97,7 +97,6 @@ class riscv_pmp_cfg extends uvm_object;
 
   constraint grain_addr_mode_c {
     foreach (pmp_cfg[i]) {
-      (pmp_granularity == 0) -> (pmp_cfg[i].a != NAPOT);
       (pmp_granularity >= 1) -> (pmp_cfg[i].a != NA4);
     }
   }
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -61,6 +61,11 @@ class riscv_pmp_cfg extends uvm_object;
   // Can be manually configured from the command line.
   bit [XLEN - 1 : 0] pmp_max_offset = {XLEN{1'b1}};
 
+  // Value to hold the end signature address to that signals the end to the test environment.
+  // Currently the design assumes that the end signature is address is equal to the signature
+  // address minus 4 Bytes.
+  bit [XLEN - 1 : 0] end_signature_addr;
+
   // used to parse addr_mode configuration from cmdline
   typedef uvm_enum_wrapper#(pmp_addr_mode_t) addr_mode_wrapper;
   pmp_addr_mode_t addr_mode;
@@ -145,7 +150,8 @@ class riscv_pmp_cfg extends uvm_object;
     pmp_cfg_already_configured = new[pmp_num_regions];
   endfunction
 
-  function void initialize(bit require_signature_addr);
+  function void initialize(bit [XLEN - 1 : 0] signature_addr);
+    end_signature_addr = signature_addr - 'h4;
     if (!pmp_randomize) begin
       set_defaults();
       setup_pmp();
@@ -316,9 +322,10 @@ class riscv_pmp_cfg extends uvm_object;
   // - If MMWP, set the config of code region to LXWR=0100 and TOR.
   // - If MML or MMWP, set requested MSECCFG with RLB hardcoded to 1.
   // - Don't override code region config if corresponding `+pmp_region_` is passed.
-  // - Set default code region config in MML mode to shared execute.
-  // - Set all other addresses and configs.
+  // - If MML, set default code region config to shared execute.
+  // - If MML or MMWP, set stack and signature regions to shared read/write.
   // - Set requested MSECCFG (including RLB).
+  // - Set all other addresses and configs.
   function void gen_pmp_instr(riscv_reg_t scratch_reg[2], ref string instr[$]);
     bit [XLEN - 1 : 0] pmp_word;
     bit [XLEN - 1 : 0] cfg_bitmask;
@@ -337,9 +344,10 @@ class riscv_pmp_cfg extends uvm_object;
         cfg_byte = {1'b1, 1'b0, 1'b0};
         instr.push_back($sformatf("csrwi 0x%0x, %0d", MSECCFG, cfg_byte));
 
-        if(pmp_randomize) begin
-          // Randomly select a PMP region to contain the code for permitting execution.
-          code_entry = $urandom_range(pmp_num_regions - 1);
+        if (pmp_randomize) begin
+          // Randomly select a PMP region to contain the code for permitting execution and two
+          // extra regions to contain the stack and the signature address.
+          code_entry = $urandom_range(pmp_num_regions - 3);
           // In case of full randomization we actually want the code region to cover main as well.
           pmp_cfg[code_entry].offset = pmp_max_offset;
         end else begin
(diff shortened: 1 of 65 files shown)


marnovandermaas:
@ctopal can you test this pull request and see whether it fixes the coverage issues you were seeing?

ctopal:
This looks good for overall improvement to how we randomize things. I tested it with Ibex and seems to be generally working, thanks for the work!

src/riscv_pmp_cfg.sv
```diff
@@ -294,51 +294,109 @@ class riscv_pmp_cfg extends uvm_object;
...
+          pmp_cfg[code_entry].offset = pmp_word;
         end
-        // Enable the selected config on region 0.
-        `uvm_info(`gfn, $sformatf("temporary pmp_word: 0x%0x", cfg_byte), UVM_DEBUG)
-        instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg[0], cfg_byte));
-        instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmpcfg_addr, scratch_reg[0]));
+        // Enable the selected config on region code_entry.
+        cfg_bitmask = cfg_byte << ((code_entry % cfg_per_csr) * 8);
+        `uvm_info(`gfn, $sformatf("temporary pmp_word: 0x%0x", cfg_bitmask), UVM_DEBUG)
+        instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg[0], cfg_bitmask));
```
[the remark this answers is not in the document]
marnovandermaas: This should now be solved.

marnovandermaas:
@vogelpi Would you mind having a look at this PR?

src/riscv_pmp_cfg.sv
```diff
@@ -294,51 +294,117 @@ class riscv_pmp_cfg extends uvm_object;
...
+          // Load _start into PMP address of previous entry to complete TOR region.
+          instr.push_back($sformatf("la x%0d, _start", scratch_reg[0]));
+          instr.push_back($sformatf("srli x%0d, x%0d, 2", scratch_reg[0], scratch_reg[0]));
+          instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmp_addr + code_entry - 1,
+                                    scratch_reg[0]));
+          `uvm_info(`gfn, $sformatf("Offset of pmp_addr_%d to _start", code_entry - 1), UVM_LOW)
+          if (!inst.get_arg_value($sformatf("+pmp_region_%d=", code_entry - 1), arg_value)) begin
+            // Currently there is no way to know the value of _start here and address 0 gets
+            // ignored below.
+            pmp_cfg[code_entry - 1].addr = 1;
```
vogelpi, line 354: Are you saying that we need to set an entry to get the TOR region from [0, main + offset] but address 0 will not be configured, so we just take next address?

Ignoring address 0 doesn't sound good to me but I see this is not your decision. And I don't expect you to fix it in this PR of course. But I would change the wording here to something like `address 0 gets currently ignored below` and On line 436 add a comment saying that we currently assume that address 0 means the corresponding entry in the config isn't set up and we ignore it. Eventually, also this should be changed.

src/riscv_pmp_cfg.sv
```diff
@@ -294,51 +294,117 @@ class riscv_pmp_cfg extends uvm_object;
...
+          tmp_pmp_cfg.a = TOR;
+          tmp_pmp_cfg.x = 1'b0;
+          tmp_pmp_cfg.w = 1'b1;
+          tmp_pmp_cfg.r = 1'b0;
           // This configuration needs to be executable in M-mode both before and after writing to
           // MSECCFG. It will deny execution for U-Mode, but this is necessary because RWX=111 in
           // MML means read only, and RW=01 is not allowed before MML is enabled.
-          cfg_byte = {1'b1, pmp_cfg[0].zero, TOR, 1'b1, 1'b0, 1'b0};
+          cfg_byte = {tmp_pmp_cfg.l, tmp_pmp_cfg.zero, tmp_pmp_cfg.a, 1'b1,
+                      1'b0, tmp_pmp_cfg.r};
```
vogelpi, line 373: Is there a reason not using `tmp_pmp_cfg.w` and `x` here after setting it up above? Also, please note that `x` and `w` got mixed up here. In the register we have from MSB to LSB: {l, zero, a, x, w, r}.

src/riscv_pmp_cfg.sv
```diff
@@ -294,51 +294,117 @@ class riscv_pmp_cfg extends uvm_object;
...
-        instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg[0], cfg_byte));
-        instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmpcfg_addr, scratch_reg[0]));
+        // Disable all configs.
+        cfg_bitmask = 0;
+        for (int i = 0; i < (pmp_num_regions / cfg_per_csr); i++) begin
+          if (i != (code_entry / cfg_per_csr)) begin
+            instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg[0], cfg_bitmask));
+            instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmpcfg_addr + i, scratch_reg[0]));
+          end
+        end
```
vogelpi, line 401: Out of curiosity, why is this needed?

vogelpi: Ah, I see you explained this in the previous conversation below. Well, I would say this is worth a comment in the code :-)

src/riscv_pmp_cfg.sv
```diff
@@ -294,51 +294,117 @@ class riscv_pmp_cfg extends uvm_object;
...
-  // - Set default region 0 config in MML mode to shared execute
-  // - Set all other addresses and configs
-  // - Set requested MSECCFG (including RLB)
+  // - If randomization, generate code region, otherwise select region 0.
+  // - Set address of code region before setting MSECCFG.
+  // - If  MML, initially set MSECCFG to MML=0, MMWP=0, RLB=1.
+  // - If  MML, set the config of code region to LXWR=1100 and TOR.
+  // - If MMWP, set the config of code region to LXWR=0100 and TOR.
+  // - If MML or MMWP, set requested MSECCFG with RLB hardcoded to 1.
+  // - Don't override code region config if corresponding `+pmp_region_` is passed.
```
vogelpi, line 303: AFAIK, this is no longer the case. If a config is passed for the code region, we override it and I think this is fine.

vogelpi (commented):
Hi @marnovandermaas ,
thanks for the PR. I had a look. It mostly looks good but I believe there is one big issue. Namely, if ePMP support is not enabled, the code region isn't set up at all now. So this would break PMP DV for any configurations without ePMP support. Can you please have a look and ensure things work both with and without ePMP support?

src/riscv_pmp_cfg.sv
```diff
@@ -294,51 +294,117 @@ class riscv_pmp_cfg extends uvm_object;
...
+          instr.push_back($sformatf("srli x%0d, x%0d, 2", scratch_reg[0], scratch_reg[0]));
+          instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmp_addr + code_entry - 1,
+                                    scratch_reg[0]));
+          `uvm_info(`gfn, $sformatf("Offset of pmp_addr_%d to _start", code_entry - 1), UVM_LOW)
+          if (!inst.get_arg_value($sformatf("+pmp_region_%d=", code_entry - 1), arg_value)) begin
+            // Currently there is no way to know the value of _start here and address 0 gets
+            // ignored below.
+            pmp_cfg[code_entry - 1].addr = 1;
+          end
+        end
```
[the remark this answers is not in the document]
marnovandermaas: I could do this for cases where ePMP is not supported, but I don't think it is necessary. The reason why we need a PMP entry for code when MML and MMWP are enabled is because M-mode access is denied if no entry exists in the PMP. When neither of these two are enabled, M-mode has full access to memory that has no PMP entry. So when PMP randomization is enabled and ePMP is disabled, in most cases the code will not be covered by any entry, which will allow execution. And in cases where code is covered by a PMP region, we want to test this behavior as well.

src/riscv_pmp_cfg.sv
marnovandermaas: I've added the two comments that you suggested.

src/riscv_pmp_cfg.sv
marnovandermaas: Good point, I added a comment explaining this.

src/riscv_pmp_cfg.sv
marnovandermaas: I think it's still true because I check that the `+pmp_region` for the code entry is not set before overwriting the pmp config here:
https://github.com/marnovandermaas/riscv-dv/blob/335465f984f41e2ce000445c95475bd99adc073c/src/riscv_pmp_cfg.sv#L386
I can change it to not check this, but I thought from a programmer's perspective the `+pmp_region` should override `+pmp_randomize`, but I can change that priority if you think that's better.

src/riscv_pmp_cfg.sv
marnovandermaas: Yes, I am overriding the `tmp_pmp_cfg` for `x` and `w` on purpose. This is because I want `xwr=100` before MML is enabled  because it is executable both before and after MML is enabled. Afterwards I want to set the config to `xwr=010` because that is both executable by M and U mode.

marnovandermaas: I tried to explain it in a comment in the code, but maybe this needs clarification. This is what it says in the code right now:
```
// This configuration needs to be executable in M-mode both before and after writing to
// MSECCFG. It will deny execution for U-Mode, but this is necessary because RWX=111 in
// MML means read only, and RW=01 is not allowed before MML is enabled.
```

src/riscv_pmp_cfg.sv
marnovandermaas: I've also added a commit to this PR allowing specifying addresses that are 0

src/riscv_pmp_cfg.sv
```diff
@@ -138,9 +143,12 @@ class riscv_pmp_cfg extends uvm_object;
     get_hex_arg_value("+pmp_max_offset=", pmp_max_offset);
     `uvm_info(`gfn, $sformatf("pmp max offset: 0x%0x", pmp_max_offset), UVM_LOW)
     pmp_cfg = new[pmp_num_regions];
+    pmp_cfg_addr_valid = new[pmp_num_regions];
+    pmp_cfg_already_configured = new[pmp_num_regions];
   endfunction
 
-  function void initialize(bit require_signature_addr);
+  function void initialize(bit [XLEN - 1 : 0] signature_addr);
+    sig_addr = signature_addr - 'h4;
```
ctopal, line 151: Maybe renaming this to `end_signature_addr` would be better for readability?

src/riscv_pmp_cfg.sv
```diff
@@ -294,52 +312,177 @@ class riscv_pmp_cfg extends uvm_object;
...
+                           tmp_pmp_cfg.x, tmp_pmp_cfg.w,    tmp_pmp_cfg.r};
+        end
+        // In case code entry is not specified in the arguments, overwrite it in pmp_cfg.
+        if (!inst.get_arg_value($sformatf("+pmp_region_%d=", code_entry), arg_value)) begin
+          pmp_cfg[code_entry].l      = tmp_pmp_cfg.l;
+          pmp_cfg[code_entry].a      = tmp_pmp_cfg.a;
+          pmp_cfg[code_entry].x      = tmp_pmp_cfg.x;
+          pmp_cfg[code_entry].w      = tmp_pmp_cfg.w;
+          pmp_cfg[code_entry].r      = tmp_pmp_cfg.r;
+        end
```
ctopal, line 410: In relation to the comments above with Pirmin, maybe consider adding a comment here. Something like "Set code_entry config to xwr=010 because that is both executable by M and U mode"

src/riscv_pmp_cfg.sv
```diff
@@ -294,52 +312,177 @@ class riscv_pmp_cfg extends uvm_object;
...
+
+        // Load the address of the kernel_stack_end into PMP stack entry.
+        instr.push_back($sformatf("la x%0d, kernel_stack_end", scratch_reg[0]));
+        instr.push_back($sformatf("srli x%0d, x%0d, 2", scratch_reg[0], scratch_reg[0]));
+        instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmp_addr + code_entry + 1,
+                                  scratch_reg[0]));
+        `uvm_info(`gfn, $sformatf("Address of pmp_addr_%d is kernel_stack_end", code_entry + 1),
+                  UVM_LOW)
+        pmp_cfg_already_configured[code_entry + 1] = 1'b1;
+        // In case stack entry is not specified in the arguments, overwrite it in pmp_cfg.
```
ctopal, line 437: Since we are not actually setting code_entry by ourselves in random case, it's not necessarily true to say this condition checks for stack entry being specified. However, I can't come up with a better solution :) Maybe you can just change the comment in order to prevent confusion?

src/riscv_pmp_cfg.sv
```diff
@@ -294,52 +312,177 @@ class riscv_pmp_cfg extends uvm_object;
...
+        // Load the signature address into PMP signature entry. This assumes the
+        // end_signature_addr = signature_addr - 4. And that both are 4 Bytes.
+        instr.push_back($sformatf("li x%0d, 0x%0x", scratch_reg[0], sig_addr));
+        instr.push_back($sformatf("srli x%0d, x%0d, 2", scratch_reg[0], scratch_reg[0]));
+        instr.push_back($sformatf("csrw 0x%0x, x%0d", base_pmp_addr + code_entry + 2,
+                                  scratch_reg[0]));
+        `uvm_info(`gfn, $sformatf("Address of pmp_addr_%d is signature_addr", code_entry + 2),
+                  UVM_LOW)
+        pmp_cfg_already_configured[code_entry + 2] = 1'b1;
+        // In case signature entry is not specified in the arguments, overwrite it in pmp_cfg.
```
ctopal, line 465: Same thing as above.

ctopal (approved):
LGTM, I made some suggestions around commenting things. Thanks a lot for the work, even reviewing it was hard!

src/riscv_pmp_cfg.sv
marnovandermaas: Thank you, this should now be changed.

src/riscv_pmp_cfg.sv
marnovandermaas: Thanks, I added a comment here talking about the LXWR value being 1010.

src/riscv_pmp_cfg.sv
marnovandermaas: I've changed this to hopefully be less confusing.

src/riscv_pmp_cfg.sv
marnovandermaas: Thanks, same resolution as above.

vogelpi (approved):
LGTM, thanks for the many clarifications @marnovandermaas !

marnovandermaas:
@weicaiyang Can you have a look at this?

weicaiyang (approved):

PR #887, opened 2022-08-03
Csr refactor

GregAC:
@ctopal @marnovandermaas @vogelpi PTAL


MikeOpenHWGroup:
FWIW I applaud the goals of this PR.  Probably would have saved OpenHW some effort writing manual test-programs.

marnovandermaas (approved):
LGTM, thanks.

ctopal (approved):
LGTM, tested it within Ibex repository. Works as expected, thanks! 

Minor nit: The knob for generating writes to read-only CSRs maybe could be named `gen_csr_ro_writes`? I suggest it because an invalid write can happen because of some other reason as well.

GregAC:
> The knob for generating writes to read-only CSRs maybe could be named gen_csr_ro_writes? I suggest it because an invalid write can happen because of some other reason as well.

Good suggestion I've made the change.

weicaiyang (approved):

PR #888, opened 2022-08-11
[pmp] No PMP exception handler when no PMP support

GregAC:
@marnovandermaas @ctopal PTAL

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -62,6 +62,11 @@ class riscv_asm_program_gen extends uvm_object;
 
   // This is the main function to generate all sections of the program.
   virtual function void gen_program();
+    // Prevent generation of PMP exception handling code where PMP is not supported
+    if (!support_pmp) begin
+      cfg.pmp_cfg.enable_pmp_exception_handler = 1'b0;
+    end
+
     instr_stream.delete();
     // Generate program header
     gen_program_header();


src/riscv_asm_program_gen.sv
```diff
@@ -62,6 +62,11 @@ class riscv_asm_program_gen extends uvm_object;
 
   // This is the main function to generate all sections of the program.
   virtual function void gen_program();
+    // Prevent generation of PMP exception handling code where PMP is not supported
+    if (!support_pmp) begin
+      cfg.pmp_cfg.enable_pmp_exception_handler = 1'b0;
+    end
+
```
ctopal, line 69: Minor nit: maybe we can have this assignment in line 952? Because we already have a clause for not having support for PMP there and that is the only place where we have functions that use `enable_pmp_exception_handler` as well.

ctopal (approved):
LGTM with a very minor nit

src/riscv_asm_program_gen.sv
GregAC: I think I prefer it here. Currently if we put it in `gen_all_trap_handler` as you suggest it will work fine but it's more susceptible to other changes meaning we end up generating some exception handler code before we get to the config change.

GregAC:
@weicaiyang PTAL

weicaiyang (approved):

PR #889, opened 2022-08-16
Tweak CSR constraints for more even read/write distribution

GregAC:
Where contraints allow a CSR to be both read and written read-only vs
read-write (and write-only) operation was effectively chosen by which
register or immediate was randomly chosen. This made read-only accesses
to such CSRs rare.

Now we explicitly choose to do a read or a read-write with an explicit
random bit that is solved before the register or immediate.

Even with this change the distruibution skews to read-write operation as
the CSRRW and CSRRWI instructions cannot be read-only and that is chosen
independently of the explicit read-only/read-write decision introduced
by this commit. Fixing this so we get a 50/50 distribution of
read-only/read-write is significantly more complex but isn't required as
the read-only case is now sufficiently common.

This also allows the removal of the macro used to define the expression
for a writeable CSR giving neater code.

### DIFF
--- a/src/isa/riscv_csr_instr.sv
+++ b/src/isa/riscv_csr_instr.sv
@@ -14,17 +14,6 @@
  * limitations under the License.
  */
 
-// This expression is needed in multiple constraints so is expressed as a macro. It could be
-// implemented within a function but the constraint solver is unable to work with function contents.
-// So when used in a constraint it will just try different values until it finds something that
-// satisifies what it requires or it gives up. When the set of writable CSRs is small relative to
-// the set of readable CSRs such failures are common. Instead by using an expression that can be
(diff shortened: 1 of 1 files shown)


GregAC:
@ctopal @marnovandermaas @hcallahan-lowrisc PTAL

ctopal (approved):
LGTM. Question: if we want to generate writes for a custom CSR, we need to provide a list of all CSRs (that we want to randomly generate writes to) including the custom ones right?

GregAC:
Yes, there's a list of supported CSRs supplied in the core setting file. On top of that there is a list of CSRs that can be written to, which you can add and remove things from via plusargs or through modifying the list before the test begins.

GregAC:
@weicaiyang PTAL

marnovandermaas (approved):
LGTM

PR #894, opened 2022-08-24
for issue 807 modify the logic

mukesh891:
please update the following lines.

Thanks


MikeOpenHWGroup:
Hi @mukesh891, I would like to recommend to the code owners that this pull-request be closed without merging.  My rationale is:

1. You have not responded to the request to sign the CLA for 3.5 years.
2. Changing from `-mcmodel=medany` to `-mcmodel=medlow` offers no real advantage.  Both limit the addressable space to +/-2G, but medlow must start at 0x0.  AFAIK, medany can also start at 0x0, so there is no disadvantage.

PR #896, opened 2022-08-31
[pmp] Add knob to suppress PMP setup code

GregAC:
By default PMP setup is included in every test which sets up a full set
of PMP regions. A test may wish to have full access to all of memory to
ensure there are no PMP failures.  This knob suppresses the general PMP
setup generation and instead generates a setup with a single region
giving full access to all of memory.

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -39,6 +39,10 @@ class riscv_pmp_cfg extends uvm_object;
   // allowing all access restrictions to be enforced.
   bit enable_pmp_exception_handler = 1'b1;
 
+  // Don't generate the usual PMP setup section, instead generate a setup that provides a single
+  // region allowing full access to all of memory from both U mode and M mode.
+  bit suppress_pmp_setup = 0;
+
   // Setting this bit to 1'b1 enables generation of the directed stream of instructions to test
   // write accesses to all supported pmpaddr[i] CSRs.
   bit enable_write_pmp_csr;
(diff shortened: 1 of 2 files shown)


GregAC:
@marnovandermaas PTAL

src/riscv_pmp_cfg.sv
```diff
@@ -307,6 +312,15 @@ class riscv_pmp_cfg extends uvm_object;
...
   endfunction
 
+  // Generates code to setup a single PMP region allowing full access to all memory
+  function void gen_pmp_enable_all(riscv_reg_t scratch_reg, ref string instr[$]);
+    // Setup region 0 to NAPOT covering the whole 32-bit address space, with RWX permissions and no
+    // lock.
+    instr.push_back($sformatf("li x%0d, 0x1fffffff", scratch_reg));
+    instr.push_back($sformatf("csrw 0x%0x, x%0d", PMPADDR0, scratch_reg));
+    instr.push_back($sformatf("csrw 0x%0x, 0x1f", PMPCFG0));
+  endfunction
```
marnovandermaas, line 322: This looks right to me:
- Three zeros on the op of the address means a 32-byte region starting from 0.
- 0x1f config byte means L=0, A=NAPOT and XWR=111.

To be consistent with the rest of the code, you can consider using TOR instead, but this is fine as well.

marnovandermaas (approved):
LGTM, thanks!

src/riscv_pmp_cfg.sv
GregAC: I did consider using TOR but preferred the simplicity of a single address register.

GregAC:
@weicaiyang PTAL

weicaiyang (approved):

PR #900, opened 2022-09-28
Add plusarg to enable ECALL in main randomized body

hcallahan-lowrisc:
Adding ECALL to the main randomized body allows us to get coverage over this instruction, but requires you to rethink how the test_done mechanism works. Currently the ECALL handler is responsible for cleanup and jumping to the end-of-test loop, so if you don't change this mechanism the test will end upon reaching the first ECALL instruction.
e.g. override the gen_test_done method in combination with enabling this option.

@ctopal @GregAC

### DIFF
--- a/src/isa/riscv_instr.sv
+++ b/src/isa/riscv_instr.sv
@@ -161,6 +161,9 @@ class riscv_instr extends uvm_object;
         end
       end
     end
+    if (!cfg.no_ecall) begin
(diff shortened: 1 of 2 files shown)


GregAC (approved):

GregAC:
@weicaiyang PTAL

ctopal (approved):

weicaiyang (approved):

PR #903, opened 2022-10-13
[pmp] Fix ePMP related trap handling

marnovandermaas:
The RISCV-DV PMP trap handler is quite an intricate procedure. In essence, it iterates through the PMP entries to find the one that caused the trap, then tries to amend that entry to give it the permissions it needs and continues the test program if that was successful. This works quite well when most entries are unlocked since whenever you get an access fault, the entry is changed to allow the operation and the instruction is retried. However, there are a few problems with this approach:
- The checking for locked entries did not take into account that MML mode.
- In cases where most entries are locked (e.g. MML mode), tests would be very short ending after just the first failed load, store or fetch.

Besides the trap handling, the population of entries had some shortcomings,  which lead to poor coverage:
- NAPOT regions would have a random address, which made it very unlikely large regions are selected, which require a consecutive sequence of 1s at the end.
- The top of TOR regions was never lower than its base.

This PR addresses these issues. This is quite a big PR, but I have split it up into smaller commits to make reviewing easier. Please have a look at each commit message, but in summary these changes include:
- Fixes to make MML tests work
- PMP trap handler:
  * Skipping faulting loads and stores
  * Checking for MML and MMWP
  * Stop overwriting `mscratch`
- Improved PMP configuration and debugging
- Improved coverage of NAPOT and illegal TOR regions

### DIFF
--- a/src/riscv_pmp_cfg.sv
+++ b/src/riscv_pmp_cfg.sv
@@ -393,7 +393,7 @@ class riscv_pmp_cfg extends uvm_object;
           stack_entry = pmp_num_regions - 2;
           sig_entry = pmp_num_regions - 1;
           // This is the default offset.
-          pmp_cfg[code_entry].offset = assign_default_addr_offset(pmp_num_regions, 0);
+          pmp_cfg[code_entry].offset = assign_default_addr_offset(pmp_num_regions, code_entry);
           pmp_cfg[pmp_num_regions - 3].offset = pmp_max_offset;
         end
 
@@ -443,7 +443,7 @@ class riscv_pmp_cfg extends uvm_object;
         // In case the randomly selected code entry is not also configured in the arguments,
         // overwrite it in pmp_cfg.
         // The pmp_config has value LXWR = 1010, which means it is executable in both M and U mode.
-        if (!inst.get_arg_value($sformatf("+pmp_region_%d=", code_entry), arg_value)) begin
+        if (!inst.get_arg_value($sformatf("+pmp_region_%0d=", code_entry), arg_value)) begin
           pmp_cfg[code_entry].l      = tmp_pmp_cfg.l;
           pmp_cfg[code_entry].a      = tmp_pmp_cfg.a;
           pmp_cfg[code_entry].x      = tmp_pmp_cfg.x;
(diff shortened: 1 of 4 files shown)


marnovandermaas:
@hcallahan-lowrisc would you be able to have a look at this?

marnovandermaas:
@vogelpi, in case you have time would you be able to review this PR?

src/riscv_pmp_cfg.sv
```diff
@@ -393,7 +393,7 @@ class riscv_pmp_cfg extends uvm_object;
           stack_entry = pmp_num_regions - 2;
           sig_entry = pmp_num_regions - 1;
           // This is the default offset.
-          pmp_cfg[code_entry].offset = assign_default_addr_offset(pmp_num_regions, 0);
+          pmp_cfg[code_entry].offset = assign_default_addr_offset(pmp_num_regions, code_entry);
```
ctopal, line 396: nit: I believe this fix probably needs to go to the first commit.

ctopal (approved):
Overall it looks good, I am not sure about the skipping in the case of seeing a load/store fault solution but also can't see another way as well. Just had a minor thing for the last commit but I think it shouldn't matter that much so I think this one is good to go.

src/riscv_pmp_cfg.sv
```diff
@@ -784,36 +783,93 @@ class riscv_pmp_cfg extends uvm_object;
...
+                 $sformatf("and x%0d, x%0d, x%0d", scratch_reg[0], scratch_reg[0], scratch_reg[4]),
+                 // Check whether instruction is compressed.
+                 $sformatf("beq x%0d, x%0d, 28f", scratch_reg[0], scratch_reg[4]),
+                 $sformatf("csrr x%0d, 0x%0x", scratch_reg[0], MEPC),
+                 // Increase MEPC by 2 in case instruction is compressed.
+                 $sformatf("addi x%0d, x%0d, 2", scratch_reg[0], scratch_reg[0]),
+                 $sformatf("csrw 0x%0x, x%0d", MEPC, scratch_reg[0]),
+                 $sformatf("j 34f"),
+                 $sformatf("28: csrr x%0d, 0x%0x", scratch_reg[0], MEPC),
+                 // Increase MEPC by 4 in case instruction is compressed.
```
GregAC, line 888: compressed -> uncompressed

GregAC (approved):
One tiny nit on a comment but otherwise this looks good. Probably not worth the faff in fixing the comment as marno is away. @hcallahan-lowrisc have you seen any issues with this, are you happy with it? If there's no problems we should get this merged.

hcallahan-lowrisc:
@GregAC @ctopal no I have not seen any issues, happy to merge as-is.

GregAC:
@weicaiyang PTAL

weicaiyang (approved):

PR #909, opened 2022-11-09
Update metrics-regress.yml

pradheepkaliraj:
Fix for "Please update the following actions to use Node.js 16: actions/checkout, actions/checkout"

### DIFF
--- a/.github/workflows/metrics-regress.yml
+++ b/.github/workflows/metrics-regress.yml
@@ -16,7 +16,7 @@ jobs:
   metrics-regression:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - run: ./scripts/metrics-regress.py $METRICS_REGRESSION_NAME $METRICS_PROJECT_ID
         env:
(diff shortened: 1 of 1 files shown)


MikeOpenHWGroup:
Are Metrics CI regressions still running?

pradheepkaliraj:
No, there is an issue with metric_regress.py when making a http request. 
It fails with the following error socket.gaierror: [Errno -3] Temporary failure in name resolution.
https://github.com/google/riscv-dv/blob/1246c391cbbc3e161178dcb1aa88f34c84d123c2/scripts/metrics-regress.py

I am looking into this.

MikeOpenHWGroup:
It is my understanding that Metrics has depreciated their cloud-based regression platform.

pradheepkaliraj:
That makes sense. The server doesn't exist anymore.  Should I revert this action then, so people don't get a regression failure mail, every time they commit a change?

MikeOpenHWGroup:
Yes, that makes sense.  Metrics has changed their product offering and it is not clear if the regression service will be offered in the near to mid-term future.

MikeOpenHWGroup:
Hi @pradheepkaliraj, this PR has stalled out.  As we discussed, please revert this action so people don't get a regression failure mail every time they commit a change.

PR #910, opened 2022-11-11
Fix bitmanip cov

GregAC:


GregAC:
@vogelpi PTAL

vogelpi (approved):
What I am wondering is whether coverage was always broken for those ZBB instructions even back when ZBB (v.0.93 and earlier) was entirely handled by `riscv_b_instr.sv`. Anyway, this looks good, thanks @GregAC !

vogelpi:
I think @weicaiyang usually merges these PRs.

weicaiyang (approved):

PR #911, opened 2022-11-15
[pmp] Add option to constrain addresses to stay in 32-bit space

GregAC:
PMP supports up to 35 bits of physical address space when XLEN==32. For cores where this higher address space doesn't exist there is limited value in testing PMP settings that use it. This introduces a way to bias PMP configs so that the majority stay within the 32-bit space.


GregAC:
@marnovandermaas PTAL

marnovandermaas (approved):
LGTM, great work Greg!

GregAC:
@weicaiyang PTAL

weicaiyang (approved):

PR #913, opened 2022-11-16
[pmp] Ensure MML PMP configurations don't dominate.

GregAC:
A constraint prevents MML being disabled when region configurations with W == 1 and R == 0 exist (this combination is reserved when MML is disabled).  This has the effect of making configurations with MML disabled rare as many random configurations will have at least one region with W == 1 and R == 0.

Add a solve before constraint so MML is decided upon before the W and R are randomized to more evenly generate MML and non-MML configurations.


GregAC:
@marnovandermaas PTAL

weicaiyang (approved):

PR #921, opened 2023-01-30
Add new extensions Zcb, Zfh, Zbkb, partial Zfa

JJ-Gaisler:
I have added a few new extensions to RISC-DV, I would appreciate if someone could take a look at the major changes and see if they disagree with any of the decisions. 

The major changes include: 

+Zcb (code reduction):
since no gcc support is in the mainline assembler yet these instructions are injected as raw data. Coverage class has been extended with these instructions and the illegal instruction class should not be able to generate legal Zcb instruction if the extension is enabled.

+Zfh (half-precision floats):
Similarly to single and double precission, RISC-V now support half-precission. Most things are identical to the F & D extension which is why the pre-existing floating_point class was extended rather than a separate class. Inital coverage support has also been added. FMV_X_H is temporarily excluded due to a bug in OVPsim.

+Zbkb (crypto):
The complementary instruction not defined in Zbb has been added to the instruction generator. This required a few instructions which were previously defined in the non-ratified bitmanip instruction class to be moved to Zbkb. I believe this makes more sense since Zbkb is ratified. Inital coverage has also been added for these instructions.

+Zfa:
Instructions has been defined for Zfa, more work will follow once OVPsim supports this extension.

The illegal instruction class has seen the following bug fixes: 1) Not all legal op-codes were considered for compressed floating instructions leading to legal instructions being injected in undesired locations (sometimes overwriting the stack-pointer since no reserved gpr check was performed).
2) ldsp and lqsp are only reserved for some XLENs.

ovpsim_log_to_trace:
Fixed a bug where some CSRs were not considered CSRs but GPRs

run.py:
Fixed bug in the regex replace, not all C's in the march string should be replaced since that breaks all zc... extension substrings.

load_store_instr_lib:
Added the Zcb load and store instructions to this class.

riscv_instr_pkg:
Added new extensions and instructions.


scottj97:
I'm only a user of this project and not a contributor or even a regular reviewer, so take this with a grain of salt.

When the change is presented as one giant 1400+ line commit, it is difficult to understand, difficult to review, difficult to merge with other changes, etc.

Consider instead using Git best practices and building the PR as a sequence of commits. I can easily see this being a dozen commits or more.

* [Writing Reviewable Code](https://secure.phabricator.com/book/phabflavor/article/writing_reviewable_code/)
* [Understanding the Git Workflow](https://sandofsky.com/workflow/git-workflow/)

src/isa/riscv_instr.sv
```diff
@@ -109,11 +109,16 @@ class riscv_instr extends uvm_object;
       end
       if (!cfg.enable_sfence && instr_name == SFENCE_VMA) continue;
       if (cfg.no_fence && (instr_name inside {FENCE, FENCE_I, SFENCE_VMA})) continue;
+      if (!(RV32Q inside {supported_isa}) && instr_name inside {FCVT_H_Q, FCVT_Q_H}) continue;
+      if (!cfg.enable_zfh_extension && instr_inst.group inside {RV32ZFH, RV64ZFH}) continue;
+      // TODO: There is a confirmd bug in the Jan 2023 version OVPsim where sign extension is
+      // not handled correctly, remove the FMV_X_H exclusion once that is fixed.
+      if (instr_name == FMV_X_H) continue;
```
scottj97, line 116: This seems inappropriate for the master branch of this project. OVPsim bugs are not our problem. If a user of this tool needs to disable it because they need OVPsim to work then they can add this in their private fork.

At least, it should be a separate commit on a separate branch `ovpsim-bug-workaround`.

scottj97:
Suggested new PR title: "Add new extensions Zcb, Zfh, Zbkb, partial Zfa"

src/isa/riscv_instr.sv
JJ-Gaisler: Fair point, I have removed the lines from the PR.

JJ-Gaisler:
> I'm only a user of this project and not a contributor or even a regular reviewer, so take this with a grain of salt.
> 
> When the change is presented as one giant 1400+ line commit, it is difficult to understand, difficult to review, difficult to merge with other changes, etc.
> 
> Consider instead using Git best practices and building the PR as a sequence of commits. I can easily see this being a dozen commits or more.
> 
>     * [Writing Reviewable Code](https://secure.phabricator.com/book/phabflavor/article/writing_reviewable_code/)
> 
>     * [Understanding the Git Workflow](https://sandofsky.com/workflow/git-workflow/)

Hi! I am a complete novice when it comes to PRs and I realize this may not have been the ideal PR in retrospect. I do unfortunately not have any intermediate commits from when I was developing this, which is why it ended up being one large commit. I could however make an effort to try to separate the PR into more than one commit, most extensions and bug-fixes should be feasible to put into different commits. I could do that if that makes things on the merging end much easier? What do other reviewers think? Is this PR acceptable as is or would you prefer if I with some delay re-submit the PR?

scottj97:
Also it's preferable to squash bugfixes (such as [030c3e8](https://github.com/chipsalliance/riscv-dv/pull/921/commits/030c3e8d4d8440895cd09ee48b04d0fa1d5d326b) appears to be) back into the original commits (using `git rebase --interactive` followed by `git push --force-with-lease`) so that the bug never appears in the history.

yaml/simulator.yaml
```diff
@@ -108,23 +108,26 @@
...
-        +incdir+<user_extension>
-        -f <cwd>/files.f
-        <cmp_opts>"
-  sim:
-    cmd: >
-      vsim -c <sim_opts> -sv_seed <seed> <cov_opts> -do <cwd>/riviera_sim.tcl
-    cov_opts: >
-      -acdb_file <out>/cov.acdb
+#RIVIERA REMOVED BECAUSE RANDOMIZATION ISSUE
+#   ADDITIONALLY, 10X SLOWER THAN VCS
```
MikeOpenHWGroup, line 112: Recommend against making vendor-specific comments.

JJ-Gaisler: Fair enough, I suppose we should try to resubmit the pull-request where we remove things that are specific to our design. I forgot we had an open pull request, so perhaps we can reorganize things a bit and move the Gaisler specific changes to another branch and then keep new extensions plus bugfixes in the branch we try to merge from.

For this particular comment we can take another look to see if we can get Riviera working, we are having issues with Riviera only randomizing the operands once per instruction, this behavior is not something we reproduce in VCS.

MikeOpenHWGroup: Better yet, let's get Aldec looking into this.   I have some contacts there.  If you agree, I'll reach out to them about this issue.

PR #931, opened 2023-04-07
Fix RV32 support

mkurc-ant:
Currently it is impossible to compile the eUVM-based instruction generator for RV32 due to errors in the code. This PR fixes those errors.


tmichalak:
@pradheepkaliraj could you take a look and let us know if there is something to be done here or if it is fine to be merged?

tmichalak (approved):

PR #935, opened 2023-05-18
PyFlow log trace to csv improvements and Renode ISS support

mczyz-antmicro:
This PR provides enhancements to the ISS used with PyFlow Instruction Generator:
- Fixed logs created with spike in a scenario with multiple spaces and enabled parsing CSR writes from them
- Log file names produced by run.py now follow the same convention
- Added missing PC values to VeeR-ISS (a.k.a. whisper) log to CSV converter
- Enabled Renode as an Instruction Set Simulator


tmichalak:
@pradheepkaliraj could you take a look and let us know if there is something to be done here or if it is fine to be merged?

kgugala (approved):
LGTM

PR #959, opened 2023-10-09
ISS usage fixes

mkurc-ant:
This PR fixes usage of certain ISSes:
 - Parameterized Renode memory size
 - Detection of `ECALL` instruction in Renode and writing its occurrence to the execution log
 - Enabled unaligned load/store access in Renode and Spike
 - Added config for Whisper, configured memory map to allow unaligned load/store


kgugala (approved):
LGTM

PR #960, opened 2023-10-09
Fix Ctrl-C processing & use as lib with recent python versions

cathales:
Hello,

We are using riscv-dv with two patches to verify [CVA6](https://github.com/openhwgroup/cva6/). We hope it will help everyone if those patches are made common.

1. The simulation was still running after hitting Ctrl-C. With this PR it is correctly killed. It is a continuation of work from @yanicasa fixing the same issue for timeouts.
2. We are using riscv-dv as a library with a relative path `dv.scripts`. Python 3.9 cannot find it albeit Python 3.7 had no issues finding it. To fix this, we found [this link](https://stackoverflow.com/questions/1944569/how-do-i-write-good-correct-package-init-py-files) and added an empty `__init__.py` file.

### DIFF
--- a/scripts/lib.py
+++ b/scripts/lib.py
@@ -103,6 +103,12 @@ def run_cmd(cmd, timeout_s=999, exit_on_error=1, check_return_code=True,
         debug_cmd.write(cmd)
         debug_cmd.write("\n\n")
         return
+    def killgroup(ps):
+        try:
+            os.killpg(os.getpgid(ps.pid), signal.SIGTERM)
+        except AttributeError: #killpg not available on windows
+            ps.kill()
+        sys.exit(130)
     try:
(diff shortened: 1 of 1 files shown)


mcockrell-google (approved):

JeanRochCoulon:
Hello,
Thank you to have reviewed and approved this PR. In OpenHW Group, and more precisely in CVA6 project, we are glad to use the riscv-dv project.
We try to be synchronized with this repository by declaring it as submodule. As you know the drawback to declare it as submodule is that we are dependant of it. Today we would be pleased to ugrade the riscv-dv hash, but we are waiting for this PR merge since two weeks. It would be great if you cold merge it on master. It would unlock cva6 project.

kgugala (approved):
LGTM

PR #981, opened 2024-05-29
Remove unused local variables and fix gen_callback

HuashuQ:
1. There are some unused local variables from function parameters so I remove them.
2. \_\_init\_\_ of class riscv_program has no input parameter but self
3. class riscv_asm_program_gen has no member 'callstack_gen'. It's a local variable in function 'gen_callstack'


HuashuQ:
@aneels3 Can you review this?

CVYang:
why closed the push? i think the bug is also unfixed

PR #984, opened 2024-06-28
allow specifying privilege modes for simulation

wsipak:
Currently we're using the default settings of privilege modes in ISS tools.
Whisper needs a 'u' in the ISA string for the User mode to be enabled. Spike, on the other hand, enables both Supervisor and User modes by default (it has a `--priv` arugment with the default of `msu`.)

We can make it consistent by adding an argument to select the modes to be enabled in simulation.
With these changes, the default selection is M-mode only.
The `--priv` argument allows enabling Supervisor and User mode, while Machine mode is mandatory and can't be disabled.

I would appreciate feedback on the default value for the `--priv` argument. I can see several ways to implement it here:
1. **default is `""`, which means M mode enabled, allows selecting S and U as `[su]`. (currently implemented)**
2. default is `"m"`, which means M mode enabled, allows selecting S and U as `m[su]` - in my view this is worse because M is mandatory anyway. Contrary to (1.) this doesn't use empty string for default.
3. default is `"msu"` - which means M,S,U modes are enabled. This wouldn't affect the default `spike` command line

Here's a PR in VeeR where I've proven VeeR's CI passes with these changes and `--priv u`:
https://github.com/chipsalliance/Cores-VeeR-EL2/pull/197

### DIFF
--- a/run.py
+++ b/run.py
@@ -124,13 +124,14 @@ def get_generator_cmd(simulator, simulator_yaml, cov, exp, debug_cmd):
     sys.exit(RET_FAIL)
 
 
-def parse_iss_yaml(iss, iss_yaml, isa, setting_dir, debug_cmd):
+def parse_iss_yaml(iss, iss_yaml, isa, priv, setting_dir, debug_cmd):
     """Parse ISS YAML to get the simulation command
 
     Args:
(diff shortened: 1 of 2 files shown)


wkkuna:
Closing.

PR #985, opened 2024-07-09
updated for python 3.12

5hayanB:
Renamed `imp` to `importlib` in `pygen/pygen_src/isa/riscv_instr.py`


flaviens:
Up. Indeed this would allow riscv-dv to work for python >= 3.12.

PR #988, opened 2024-07-12
run: fix disabling the `c` extension

fkokosinski:
This PR fixes the way that the `c` extension is disabled, i.e. removed from the arch ISA string. Before this change, for GCC where Zicsr and Zifencei constitute separate extensions, the following ISA string:
```
  rv32imc_zicsr_zifencei_zba_zbb_zbc_zbs
```
Would be changed to (note incorrect `_zisr` vs correct `_zicsr`):
```
  rv32im_zisr_zifenei_zba_zbb_zb_zbs
```
Instead of the expected:
```
  rv32im_zicsr_zifencei_zba_zbb_zbc_zbs
```


wkkuna:
Closing.

PR #991, opened 2024-08-28
Make renode_wrapper more configurable

MateuszKarlic:
Allow to pass more options to `renode_wrapper` such as:
* CPU model
* privilege levels
* additional CPU constructor parameters

Also this adds an additional print to `instr_trace_compare.py` and allows to pass additional arguments to ISS with `--iss_opts=` switch.

This PR depends on: https://github.com/chipsalliance/riscv-dv/pull/984 and should be considered for merge only after those changes are applied. The first commit here is taken directly from that PR.


wkkuna:
Closing.

PR #995, opened 2024-12-11
Fix mismatch in push stack logic for branch instruction generation

Bounti:
Some invalid labels are propagated to non branch instructions when generating push stack logic with cfg.enable_branch true. The issue is due to the way the get_rand_instr is called with ```.include_category({BRANCH})``` that does not seem to restrict properly the generated instructions. Workaround is to explicitly list accepted instructions. 

Reported error when compiling generated testcases:
```
illegal operands `sd s11,sub_4_stack_p(s5)'
```


Bounti:
It seems that the issue happens when the supported_isa does not support RV32I and/or RV32C while enable_branch is on. Riscv_instr is then initialized without any branch instruction, so when calling get_instr_rand  with include_category set to BRANCH non branch instruction are generated (a sanity check is also missing in this function)

PR #1028, opened 2026-02-06
imp module removed in Python 3.12

eiji-y:
For backward compatibility, using conditional imports. (sys.version_info >= (3, 12))
This commit is based on the previous pull request [updated for python 3.12#985
](https://github.com/chipsalliance/riscv-dv/pull/985).

### DIFF
--- a/pygen/pygen_src/isa/riscv_instr.py
+++ b/pygen/pygen_src/isa/riscv_instr.py
@@ -15,7 +15,10 @@ import copy
 import sys
 import random
 import vsc
-from imp import reload
+if sys.version_info >= (3, 12):
+    from importlib import reload
+else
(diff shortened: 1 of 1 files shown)


pygen/pygen_src/isa/riscv_instr.py
```diff
@@ -15,7 +15,10 @@
 import sys
 import random
 import vsc
-from imp import reload
+if sys.version_info >= (3, 12):
+    from importlib import reload
+else
```
sweeneyal: You forgot a ":" here.

eiji-y: I am sorry for not having performed the final verification of pyflow.
I updated the branch.

PR #1035, opened 2026-07-21
gen_csr_test: only read read-only CSRs to avoid illegal writes

avinashkollu-git:
Motivated by lowRISC/ibex#1337.

## Problem
`gen_csr_test.py` emits writing instructions (`csrrw`, and `csrrs`/`csrrc` with a non-zero source, plus the non-zero-immediate forms) for **every** CSR, including read-only ones. Per the RISC-V ISA, any attempt to write a read-only CSR raises an illegal instruction exception. The generated test has no exception handler, so the run hangs until timeout. This is why read-only CSRs such as `mvendorid`, `marchid`, `mimpid` and `mhartid` currently have to be left out of CSR description files.

Reproduced with a minimal read-only CSR (`mvendorid`, 0xF11):
```
	csrrw x10, 3857, x3     # write to a read-only CSR -> illegal instruction
	csrrs x10, 3857, x3
	csrrc x10, 3857, x3
	...
```

## Fix
- In `get_csr_map()`, mark a CSR as read-only when it has fields and none of them are writable.
- In `gen_csr_instr()`, for a read-only CSR emit only non-writing accesses: `csrrs`/`csrrc` with `x0` as the source register read the CSR with no write side effect. The value read back is self-checked against the CSR's masked reset value.
- Writable CSRs are untouched and keep full read/write coverage.

## After the fix
```
	# mvendorid
	csrrs x10, 3857, x0
	li x3, 0x0000002a
	bne x3, x10, csr_fail
	csrrc x10, 3857, x0
	li x3, 0x0000002a
	bne x3, x10, csr_fail
```
No illegal writes; the read value is self-checked. Verified locally: read-only CSRs emit zero writing instructions, and a writable CSR in the same run still generates its full set of write ops (no regression).

### DIFF
--- a/scripts/gen_csr_test.py
+++ b/scripts/gen_csr_test.py
@@ -101,7 +101,7 @@ def get_csr_map(csr_file, xlen):
 
     Returns:
       A dictionary contining mappings for each CSR, of the form:
-      { csr_name : [csr_address, csr_val_bitarray, csr_write_fields, csr_read_mask_bitarray] }
+      { csr_name : [csr_address, csr_val_bitarray, csr_write_fields, csr_read_mask_bitarray, csr_read_only] }
     """
     rv_string = "rv{}".format(str(xlen))
     csrs = {}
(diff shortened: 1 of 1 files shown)


mcockrell-google:
Thanks for looking into this and providing a fix.  MikeOpenHWGroup will take a look at merging once we his back.

avinashkollu-git:
Thanks @mcockrell-google. I verified the change locally: read-only CSRs now emit only non-writing accesses (no illegal writes), and writable CSRs keep full coverage, so there should be no regression. Happy to adjust if @MikeOpenHWGroup has any feedback during review. Thanks both.

MikeOpenHWGroup (approved):
LGTM @avinashkollu-git.  I note that your fix also works with CSRs that have a mix of R and WARL fields.

Hi @mcockrell-google, I do not have write permissions to this repo so I'll leave it to you for the final merge.

avinashkollu-git:
Thanks for the review and approval, Mike, and good catch on the mixed R and WARL case. Thanks @mcockrell-google, please merge whenever convenient, happy to rebase if needed.

PR #1037, opened 2026-07-31
Deprecate gen_csr_test.py

MikeOpenHWGroup:
This pull-request deprecates, but does not remove, the `gen_csr_test.py` script.  The rationale for this somewhat drastic change:
1. The CSR access-mode configurations (e.g. WPRI, WARL, etc.) are not well supported.
2. CSR testing is now extremely well support by the so-called "[ACT4](https://github.com/riscv/riscv-arch-test)" architecture certification tests supported by RVI.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -72,6 +72,7 @@ jobs:
         run: |
           yum update -y
           yum install -y python39
+          python3.9 -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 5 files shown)


mcockrell-google (approved):
Looks good, thanks!

MikeOpenHWGroup:
Thanks for the quick review @mcockrell-google.  Most of the CI checks seem to be failing due to the dreaded version collision problem, so I have pushed in a fix.

Issue #871, opened 2022-07-08
gcc_compile substitution regex issue


JJ-Gaisler:
In the gcc_compile function in run.py there is a re.sub() that removes the C extension when the option "disable_compressed_instr" is selected in the testlist.
```python
if 'gen_opts' in test:
    # Disable compressed instruction
    if re.search('disable_compressed_instr', test['gen_opts']):
        test_isa = re.sub("c", "", test_isa) #problematic line
```

I had some problems with this substitution when compiling with --isa rv64gc_zba_zbb_zbc_zbs since the second C in zbc is also removed and hence the compile fails. I suggest to add the count argument to re.sub and set it to one.

```python
if 'gen_opts' in test:
    # Disable compressed instruction
    if re.search('disable_compressed_instr', test['gen_opts']):
        test_isa = re.sub("c", "", test_isa, 1) #changed
```

PR #1039, opened 2026-08-21
Fixed regex-based removal of C extension if "disable_compressed_instr=1"

asierfdln:
If a compilation ISA string like rv32imaf_zicsr was passed with the option "disable_compressed_instr" set to "1", it would remove the "c" character from "zicsr".


asierfdln:
Updated the regex around "disable_compressed_instr" so that it admits ISA strings with extensions after C as well (B, V, etc.)

asierfdln:
It seems to address also issue #871 , which is the exact same issue I was having

Issue #999, opened 2025-01-30
Race Condition Handling for Multiple Harts


ankurcere:
Does RISC-V DV handle race conditions when running tests with multiple harts? If so, what mechanisms ensure proper synchronization (e.g., barriers, locks)? If not, what is the recommended approach to avoid unintended data corruption? Any guidance or examples would be greatly appreciated.

PR #1040, opened 2026-08-21
Multihart ecall barrier

asierfdln:
Per commit description:
```
    ecall_handler: add atomic barrier and per-hart end-of-test sequences

    Only hart 0 should jump to write_tohost to terminate simulation. Without
    synchronization, hart 0 can reach the ecall handler and exit before other
    harts finish their test programs, cutting their traces short and causing
    false mismatches in log-comparison-based verification.

    When the A extension is supported and num_of_harts > 1, all harts now
    atomically increment a shared counter (ecall_barrier_cnt) via amoadd.w
    before entering their end-of-test paths. Hart 0 spins until the counter
    reaches num_of_harts, then proceeds to write_tohost. Non-zero harts park
    in either a wfi loop or a nop loop depending on the new endtest_wfi knob.
    
    A new cfg.endtest_wfi flag (default 1) is introduced separately from the
    existing cfg.no_wfi so that disabling wfi in the random instruction stream
    does not also remove the end-of-test halt behavior.
```

### DIFF
--- a/src/riscv_asm_program_gen.sv
+++ b/src/riscv_asm_program_gen.sv
@@ -351,11 +351,12 @@ class riscv_asm_program_gen extends uvm_object;
   virtual function void gen_data_page_begin(int hart);
     bit atomic_supported = (RV32A inside {supported_isa}) ||
                            (RV64A inside {supported_isa});
-    instr_stream.push_back(".section .data");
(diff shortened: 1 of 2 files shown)


asierfdln:
This PR seems to also "address" "issue" #999 .

